PR #4, opened 2016-05-20
[WIP] Use systemd to auto-start node on boot and break into modular scripts

benhylau:
Haven't had a chance to test this, but this should automatically bring up the mesh node (Mesh Point + cjdns) after system boot, after **install.sh** is run for the first time.


benhylau:
I have tested this many times. Flash Raspbian and run:

```
wget https://raw.githubusercontent.com/tomeshnet/prototype-cjdns-pi2/autorun/scripts/install && chmod +x install && ./install
```

Your Pi 2 will reboot at the end. Connect to it again, and the **Mesh Point interface** and **cjdns** would have automatically started with boot. Use the following commands to verify that they are running:

```
pi@raspberrypi:~ $ ifconfig mesh0
mesh0     Link encap:Ethernet  HWaddr e8:94:f6:15:32:2a
          inet addr:169.254.238.175  Bcast:169.254.255.255  Mask:255.255.0.0
          inet6 addr: fe80::c955:9e6e:a069:afb9/64 Scope:Link
          UP BROADCAST RUNNING MULTICAST  MTU:1500  Metric:1
          RX packets:1788 errors:0 dropped:0 overruns:0 frame:0
          TX packets:1829 errors:0 dropped:0 overruns:0 carrier:0
          collisions:0 txqueuelen:1000
          RX bytes:270106 (263.7 KiB)  TX bytes:332427 (324.6 KiB)

pi@raspberrypi:~ $ nodejs /home/pi/cjdns/tools/peerStats
v17.0000.0000.0000.0015.q25ul3bmnpg67tt9kgb7jqssu323n8n7d101nzwkpd3hsfx14kk0.k ESTABLISHED in 2kb/s out 2kb/s  LOS 6 "outer"
v17.0000.0000.0000.0013.q25ul3bmnpg67tt9kgb7jqssu323n8n7d101nzwkpd3hsfx14kk0.k ESTABLISHED in 0kb/s out 0kb/s  LOS 6 "outer"
```

You can also check on the `systemd` services with:

```
$ sudo systemctl -l status mesh.service
$ sudo systemctl -l status cjdns.service
```

scripts/install
```diff
@@ -0,0 +1,50 @@
+#!/bin/sh
+
+set -e
+
+TAG_PROTOTYPE_CJDNS_PI2=autorun
```
benhylau: Remember to change this to **master** before merging.

scripts/install
```diff
@@ -0,0 +1,50 @@
...
+TAG_CJDNS=cjdns-v17.3
```
benhylau: @JoshuaOpolko Though the issue we encountered is fixed on crashey, let's stay on a tag so we never compile against changing code again. **v17.3** is the latest stable.

PR #8, opened 2016-06-14
Add script to check node status and current peers

benhylau:
- [x] Install and uninstall script properly without need of absolute path to **~/cjdns/tools/pingAll**
- [x] Use **peerStats** not **pingAll**

### DIFF
--- a/scripts/status
+++ b/scripts/status
@@ -25,5 +25,8 @@ echo -e 'NODE'
 sudo grep '"ipv6"' /etc/cjdroute.conf | awk '{ print $2 }' | sed 's/[",]//g'
 echo -e '---------------------------------------'
 echo -e 'PEERS'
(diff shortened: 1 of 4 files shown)


scripts/status
```diff
@@ -0,0 +1,29 @@
...
+        echo -e "cjdns Service ................ $ACTIVE"
+else
+        echo -e "cjdns Service .............. $INACTIVE"
+fi
+echo -e '---------------------------------------'
+echo -e 'NODE'
+sudo grep '"ipv6"' /etc/cjdroute.conf | awk '{ print $2 }' | sed 's/[",]//g'
+echo -e '---------------------------------------'
+echo -e 'PEERS'
+nodejs ~/cjdns/tools/pingAll | grep '@' | sed 's/@.*//'
```
benhylau: This is wrong. It pings all routes not just peers.

PR #12, opened 2016-08-14
IPFS installer

claudiobizzotto:
Add option to install IPFS with other services.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -4,6 +4,7 @@ set -e
 
 TAG_PROTOTYPE_CJDNS_PI2=master
 TAG_CJDNS=cjdns-v17.3
+GO_IPFS_VERSION=v0.4.2
 
 # Get tools
 if ! [ "$(which git)" ] || ! [ "$(which nodejs)" ] || ! [ "$(which iperf3)" ]; then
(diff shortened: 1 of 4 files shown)


scripts/install
```diff
@@ -48,5 +48,21 @@ sudo chmod 644 /lib/systemd/system/cjdns.service
 sudo systemctl daemon-reload
 sudo systemctl enable cjdns.service
 
+# IPFS
+if [ ! -x "$(command -v ipfs)" ] && [ ! -z "$WITH_IPFS" -a "$WITH_IPFS" = "true" ]; then
+    # Install and initialize IPFS
+    GO_IPFS_VERSION=v0.4.2
```
benhylau: Move this to top of file where all the other constants are?

scripts/ipfs-daemon.service
```diff
@@ -0,0 +1,19 @@
+# /etc/systemd/system/ipfs-daemon.service
```
benhylau: This shouldn't be here right? It's not even where it's installed.

claudiobizzotto:
Issues addressed, @benhylau

PR #15, opened 2016-08-28
Make installation work for the RPi 1

claudiobizzotto:
* Convert install script from sh to bash
* Make install script behave according to RPi version: for RPi 1,
  some flags have to be changed/removed when compiling CJDNS

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1,10 +1,32 @@
-#!/bin/sh
+#!/usr/bin/env bash
 
 set -e
 
 TAG_PROTOTYPE_CJDNS_PI2=master
 TAG_CJDNS=cjdns-v17.3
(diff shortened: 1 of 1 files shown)


claudiobizzotto:
**Before merging:** I've tested this on an Pi 1 and an Pi 3 today - would be nice to have someone test it on an Pi 2 as well.

garrying:
@claudiobizzotto Tested on my Pi 1 and it looks good. @benhylau has a Pi 2 to test with.

claudiobizzotto:
Thanks, @garrying :+1:

benhylau:
Code looks good. Will test on Pi 2 and 3 and do benchmarks to make sure `CFLAGS` are passed properly. Weekend :)

PR #37, opened 2016-11-26
curl for install script and pipe to bash without leaving a file

benhylau:


benhylau:
Handle `WITH_IPFS` and `WITH_WIFI_AP` as part of the script.

PR #44, opened 2017-01-01
Make install scripts modular and mesh point interface optional

benhylau:
This breaks up the monolithic install/uninstall script into modular scripts in each of the optional modules:

- Mesh Point interface _(newly made optional)_
- WiFi Access Point
- IPFS

The install script for each module can now be run standalone after initial installation. This PR also addresses https://github.com/tomeshnet/prototype-cjdns-pi2/issues/38 and part of https://github.com/tomeshnet/prototype-cjdns-pi2/issues/33.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -2,9 +2,8 @@
 
 set -e
 
-TAG_PROTOTYPE_CJDNS_PI2=master
+TAG_PROTOTYPE_CJDNS_PI2=modular
 TAG_CJDNS=cjdns-v18
-GO_IPFS_VERSION=v0.4.2
 
 # Get RPi version and set flags accordingly
 RPI_REVISION=`sed -rn 's/Revision\s+\:\s+([0-9a-z_\-\s\,\(\)]+)/\1/p' /proc/cpuinfo`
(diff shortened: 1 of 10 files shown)


scripts/install
```diff
@@ -21,13 +20,45 @@ elif [[ $RPI_REVISION == *"00"* ]]; then
...
     echo ">>> Starting installation on Raspberry Pi 3..."
 else
     echo ">>> Unknown Raspberry Pi model"
     exit 1
 fi
 
+# Prompt and set missing flags
+if [ -z "$WITH_MESH_POINT" -o "$WITH_MESH_POINT" != "true" -a "$WITH_MESH_POINT" != "false" ]; then
+    read -p "Configure Mesh Point interface (Y/n)? " CHOICE_MESH_POINT;
+    if [ "$CHOICE_MESH_POINT" == "Y" ]; then
```
garrying: - Should accept `y`, `Y`, `Yes`, `yes`
- The prompt has a capital `Y` suggesting that is the default. An empty response returns the `else` statement, which I think is incorrect.

scripts/install
```diff
@@ -80,66 +102,19 @@ sudo chmod 644 /lib/systemd/system/cjdns-resume.service
...
-# WiFi Access Point on RPi3
-if [[ $RPI_REVISION == *"a02082"* || $RPI_REVISION = *"a22082"* ]] && [ ! -z "$WITH_WIFI_AP" -a "$WITH_WIFI_AP" = "true" ]; then
-    # Install hostapd and dnsmasq to run IEEE 802.11 Access Point
-    if ! [ "$(which hostapd)" ] || ! [ "$(dnsmasq)" ]; then
-        sudo apt-get update
-        sudo apt-get install hostapd dnsmasq -y
-    fi
+# 802.11s Mesh Point interface
+if [ ! -z "$WITH_MESH_POINT" -a "$WITH_MESH_POINT" == "true" ]; then
+    source prototype-cjdns-pi2/scripts/mesh-point/install
```
garrying: install fails here for me: `prototype-cjdns-pi2/scripts/mesh-point/install: No such file or directory`

benhylau: You'd be running this off master (without the file) unless you run from the commit before this change: https://github.com/tomeshnet/prototype-cjdns-pi2/pull/44/commits/cb7df6cdcec0def11f7d3b0185185340f334d634

scripts/install
benhylau: Right, let me fix that, and in `uninstall` too. I guess in the uninstall scripts the default should be to keep the files.

scripts/ipfs/uninstall
```diff
@@ -0,0 +1,20 @@
...
+
+# Uninstall IPFS
+sudo systemctl disable ipfs-daemon.service
+sudo systemctl stop ipfs-daemon.service 2>/dev/null || true
+sudo rm -f /usr/local/bin/ipfs
+sudo rm -f /lib/systemd/system/ipfs-daemon.service
+
+if [ -d "/home/pi/.ipfs" ]; then
+    echo "Found /home/pi/.ipfs"
+    read -p "Remove your IPFS data (Y/n)? " RM_IPFS_HOME;
```
benhylau: @garrying Change this to: **Keep your IPFS data (Y/n)?**
`<empty>`, `Y`, `y`, `Yes`, `yes` all leave the file untouched, otherwise remove?

garrying: Sounds good to me

scripts/install
garrying: Got it. Works on my end.

PR #48, opened 2017-01-19
NAT66 all local IPv6 traffic to cjdns network interface

benhylau:
Allow all client devices connected to local AP to access Hyperboria resources via NAT66. This addresses https://github.com/tomeshnet/prototype-cjdns-pi2/issues/46

### DIFF
--- a/scripts/hostapd/nat.sh
+++ b/scripts/hostapd/nat.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+# Forward all IPv4 traffic from the internal network to the eth0 device and mask with the eth0 external IP address
+iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+
+# Forward all IPv6 traffic from the internal network to the tun0 device and mask with the tun0 external IP address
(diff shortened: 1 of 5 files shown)


benhylau:
@interfect thanks for your help on this.

scripts/hostapd/nat.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# Forward all IPv4 traffic from the internal network to the eth0 device and mask with the eth0 external IP address
+iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+
+# Forward all IPv6 traffic from the internal network to the tun0 device and mask with the tun0 external IP address
+ip6tables -t nat -A POSTROUTING -o tun0 -j MASQUERADE
```
benhylau: How do I just forward `fc00::/8` to `tun0` and leave everything else to `eth0`?

benhylau: >cjd | benhylau: what you have will do what you want because already tun0 will only have a route on it which routes fc00::/8
>madafoo | benhylau, traffic outoing to which interface is chose by the route table, so u do not need do anything at all.

No change needed

PR #49, opened 2017-01-19
NO_TEST when building for RPi 1

Famicoman:
Add condition to use NO_TEST var for RPi 1. Otherwise the RPi 1 fails tests on build. Apparantly this variable can be equal to 0/1 and the build process does not perform tests. Looking at cjdns code, I believe just the presence of NO_TEST is enough to ignore tests, and we don't want that for other RPi revisions.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -84,7 +84,11 @@ fi
 # Build cjdns
 if ! [ -x "/opt/cjdns/cjdroute" ]; then
     here=`pwd`
-    cd /opt/cjdns && sudo Seccomp_NO=1 NO_NEON=$CJDNS_NO_NEON CFLAGS="$CJDNS_CFLAGS" ./do && cd $here
+    if [[ $RPI_REVISION == *"00"* ]]; then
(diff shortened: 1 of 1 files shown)


scripts/install
```diff
@@ -84,7 +84,11 @@ fi
 # Build cjdns
 if ! [ -x "/opt/cjdns/cjdroute" ]; then
     here=`pwd`
-    cd /opt/cjdns && sudo Seccomp_NO=1 NO_NEON=$CJDNS_NO_NEON CFLAGS="$CJDNS_CFLAGS" ./do && cd $here
+    if [[ $RPI_REVISION == *"00"* ]]; then
```
benhylau: Can we switch up in https://github.com/Famicoman/prototype-cjdns-pi2/blob/05c76b3eb07a3fc5ae7227c296dfe4045fce1689/scripts/install#L16 where we do all other Pi revision-specific things?

Famicoman: You know, I thought it wasn't possible to do so but I have an idea now looking back. I'll close this now and work it out!

PR #52, opened 2017-01-27
NO_TEST when building for RPi 1

Famicoman:
Redid my last PR to reduce duplication in cjdns build commands.

This change sets a variable to "" that gets assigned "NO_TEST=1" if we are building for the RPI 1 as it needs tests to turn off. Simply having NO_TEST equal to anything will cause tests to be ignored, so we cant toggle it with a value of 0/1 or anything. Echoing our var seems to work beautifully whether it contains NO_TEST=1 or of it is empty.

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -9,6 +9,7 @@ TAG_CJDNS=cjdns-v18
 RPI_REVISION=`sed -rn 's/Revision\s+\:\s+([0-9a-z_\-\s\,\(\)]+)/\1/p' /proc/cpuinfo`
 CJDNS_CFLAGS="-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm"
 CJDNS_NO_NEON=0
+CJDNS_NO_TEST=""
 
 if [[ $RPI_REVISION == *"900092"* ]]; then
     echo ">>> This Raspberry Pi model (zero) is not supported"
@@ -17,6 +18,7 @@ elif [[ $RPI_REVISION == *"00"* ]]; then
     # Skip optimizations and ARM Neon when building cjdns
     CJDNS_CFLAGS="-s -static -Wall"
     CJDNS_NO_NEON=1
+    CJDNS_NO_TEST="NO_TEST=1"
     echo ">>> Starting installation on Raspberry Pi 1..."
 elif [[ $RPI_REVISION == *"a01041"* || $RPI_REVISION == *"a21041"* ]]; then
     echo ">>> Starting installation on Raspberry Pi 2..."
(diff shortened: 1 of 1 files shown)


benhylau:
@Famicoman I wonder if we can just simplify the logic to:

```
# Get RPi version and set flags accordingly
RPI_REVISION=`sed -rn 's/Revision\s+\:\s+([0-9a-z_\-\s\,\(\)]+)/\1/p' /proc/cpuinfo`
if [[ $RPI_REVISION == *"900092"* ]]; then
    echo "This Raspberry Pi model (zero) is not supported"
    exit 0
elif [[ $RPI_REVISION == *"00"* ]]; then
    echo "Starting installation on Raspberry Pi 1..."
    Seccomp_NO=1
    NO_NEON=1
    NO_TEST=1
    CFLAGS="-s -static -Wall"
elif [[ $RPI_REVISION == *"a01041"* || $RPI_REVISION == *"a21041"* ]]; then
    echo "Starting installation on Raspberry Pi 2..."
    Seccomp_NO=1
    CFLAGS="-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm"
elif [[ $RPI_REVISION == *"a02082"* || $RPI_REVISION == *"a22082"* ]]; then
    echo "Starting installation on Raspberry Pi 3..."
    Seccomp_NO=1
    CFLAGS="-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm"
else
    echo "Unknown Raspberry Pi model"
    exit 1
fi
```

Then at the bottom:

```
cd /opt/cjdns && sudo ./do && cd $here
```

@stigtsp This script just exits on a Pi Zero. What did you do to get it working?

@cjd Any idea why tests fail on a Pi 1? And @Famicoman's comment suggests there is a bug with cjdns parsing/handling of the `NO_TEST` value.

Famicoman:
@benhylau I don't think this approach works, it looks like they need to be on the same line. Run the following:

    NO_TEST=1
    Seccomp_NO=1
    ./do

You will not see the log that Seccomp is off and then get a build failure during tests (on an RPi 1).

benhylau:
Ahh you're right. Probably why I didn't do that from the beginning... Perhaps we can do this?

```
CJDNS_BUILD_FLAGS="Seccomp_NO=1 NO_NEON=1 NO_TEST=1 CFLAGS=\"-s -static -Wall\""
```

Then:

```
cd /opt/cjdns && sudo "$CJDNS_BUILD_FLAGS" ./do && cd $here
```

Just wanted to handle all the flags in one consistent way.

benhylau:
Ahh this is the output on the Pi Zero. We need to add to the script:
```
pi@raspberrypi:~ $ sed -rn 's/Revision\s+\:\s+([0-9a-z_\-\s\,\(\)]+)/\1/p' /proc/cpuinfo
900093
```

benhylau:
Merged. The last two comments re: consistency and pi zero support will be addressed in a subsequent PR. Thanks @Famicoman

PR #55, opened 2017-02-04
Things

benhylau:
This addresses all the little things that I always wanted to add. I don't have hardware with me to test at the moment, so if someone wants to try it out, this is how the new branch-specific testing is supposed to work in my mind:

```
$ wget https://raw.githubusercontent.com/tomeshnet/prototype-cjdns-pi2/things/scripts/install && chmod +x install && TAG_PROTOTYPE_CJDNS_PI2=things WITH_MESH_POINT=true WITH_WIFI_AP=true WITH_IPFS=true ./install
```

cc. @stigtsp

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -118,6 +118,15 @@ if [ ! -x "$(command -v ipfs)" ] && [ ! -z "$WITH_IPFS" -a "$WITH_IPFS" == "true
     source prototype-cjdns-pi2/scripts/ipfs/install
 fi
 
+# Install node status script
+sudo cp prototype-cjdns-pi2/scripts/status /usr/local/bin/status
+echo -e "Run \e[1;32mstatus\e[0m anytime to print the status of your node"
+
+# Print node status on login
+cp .profile .bash_profile
+echo "" >> .bash_profile
+echo "# print mesh node status" >> .bash_profile
+echo "status" >> .bash_profile
+
 # Reboot device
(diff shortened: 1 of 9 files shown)


scripts/install
```diff
@@ -2,123 +2,28 @@
...
-fi
+# Export environment variables
+export WITH_MESH_POINT
+export WITH_WIFI_AP
+export WITH_IPFS
 
-# Reboot device
-sudo reboot
+# Run the actual installation script
+./install2
```
garrying: `./install2` is in `scripts`. https://github.com/tomeshnet/prototype-cjdns-pi2/blob/things/scripts/install#L20 should probably `cd` into it?

scripts/install
```diff
@@ -17,7 +17,7 @@ fi
 if [ -z "$TAG_PROTOTYPE_CJDNS_PI2" ]; then
     TAG_PROTOTYPE_CJDNS_PI2=master
 fi
-cd prototype-cjdns-pi2
+cd prototype-cjdns-pi2/scripts
```
garrying: This breaks some of the install paths in the `install2`.

benhylau:
@garrying I feel like cd-ing into `scripts` and removing that from the paths in `install2` is cleaner?

garrying:
@benhylau Sure. I'm cool with that

PR #56, opened 2017-02-09
Remember MESH_NAME for later use

benhylau:


benhylau:
Oh we _created_ the `.bash_profile`, it will fallback to another file

PR #63, opened 2017-02-25
Change hostapd to use EAP instead of WPA-PSK

rTreutlein:
cjd wanted to have a simple way to have EAP authentication on those nodes and this enables this by default.
Things that could be changed:
ask for default UserName/Password currently "guest/cjdnsrocks"
ask the user to modify the configs for certificate generation
ask wether to use WPA-PSK or EAP

I also had to fix the dnsmasq.conf as otherwise my client was waiting for an IP forever.


cjdelisle:
This is a huge deal because normally with WPA-PSK people can intercept traffic if they know the password. With EAP (my understanding is) everyone gets their own key and so you can publish the password widely and still the traffic cannot be intercepted by random people.

benhylau:
Thanks @rTreutlein I will review and try this out after the weekend when I have my Pi.

scripts/hostapd/install
```diff
@@ -32,15 +32,18 @@ sudo sed -i 's/#net.ipv6.conf.all.forwarding=1/net.ipv6.conf.all.forwarding=1/'
...
 sudo cp "$BASE_DIR/hostapd.conf" /etc/hostapd/hostapd.conf
+sudo cp "$BASE_DIR/hostapd.eap_user" /etc/hostapd/hostapd.eap_user
+
+/bin/bash $BASE_DIR/certs/bootstrap $BASE_DIR/certs
+
+sudo cp "$BASE_DIR/certs/ca.pem" /etc/hostapd/ca.pem
+sudo cp "$BASE_DIR/certs/server.pem" /etc/hostapd/server.pem
+
 sudo echo "ssid=$APSSID" | sudo tee --append /etc/hostapd/hostapd.conf > /dev/null
-sudo echo "wpa_passphrase=$APPASS" | sudo tee --append /etc/hostapd/hostapd.conf > /dev/null
```
benhylau: Keep this and the password selection and echo them into **/etc/hostapd/hostapd.eap_user** for EAP.

scripts/hostapd/install
```diff
@@ -32,15 +32,18 @@ sudo sed -i 's/#net.ipv6.conf.all.forwarding=1/net.ipv6.conf.all.forwarding=1/'
...
 
 # Configure network with hostapd
 sudo cp "$BASE_DIR/nat.sh" /etc/hostapd/nat.sh
 sudo cp "$BASE_DIR/hostapd.conf" /etc/hostapd/hostapd.conf
+sudo cp "$BASE_DIR/hostapd.eap_user" /etc/hostapd/hostapd.eap_user
+
+/bin/bash $BASE_DIR/certs/bootstrap $BASE_DIR/certs
+
+sudo cp "$BASE_DIR/certs/ca.pem" /etc/hostapd/ca.pem
+sudo cp "$BASE_DIR/certs/server.pem" /etc/hostapd/server.pem
```
benhylau: We need to handle these properly in the **uninstall** of the module.

rTreutlein:
@benhylau The Clients don't need to install a certificate they can just choose to not validate it. Of course, that has the slight risk somebody could impersonate your AP.
I'll fix the things you suggested but probably not before next weekend.

benhylau:
Sounds good. I will try out the current code this week. I wonder how well this is supported across client devices. By the way, we are talking about how clients can roam across APs (same SSID, password, and probably we need shared certs as well?). The discussion is on https://chat.tomesh.net/#/room/#software:tomesh.net also bridged to `freenode/#tomesh-software`

scripts/hostapd/certs/ca.cnf
```diff
@@ -0,0 +1,59 @@
...
+certificate		= $dir/ca.pem
+serial			= $dir/serial
+crl			= $dir/crl.pem
+private_key		= $dir/ca.key
+RANDFILE		= $dir/.rand
+name_opt		= ca_default
+cert_opt		= ca_default
+default_days		= 365
+default_crl_days	= 30
+default_md		= md5
```
rTreutlein: Actually, since they recently broke sha1 shouldn't we use something like sha256. Not really sure if that's supported, though.

benhylau: Yea I am just worried about compatibility issues: https://github.com/tomeshnet/prototype-cjdns-pi2/pull/63/files#diff-cc31f3ada4ea58c3308ada2613b24837R213 given this is a prototype repo I think sha1 is sufficient?

scripts/hostapd/install
benhylau: @pierrebeaucamp suggested a configuration where any username and password is accepted. Maybe we can default to that if no a blank password is entered in the prompt? Something like:

```
read -p "Set WPA2-EAP password for WiFi Access Point $APSSID (8-63 characters, or blank to accept any username and password): " APPASS;
```

scripts/hostapd/certs/ca.cnf
pierrebeaucamp: I would rather use sha256 (given this is a prototype repo) and migrate down to sha1 in case we find any incompatibilities. Let's make strong security the default, not an afterthought.

cschmittiey:
I'm a couple weeks late to the party but I agree, let's start off with sha256.

rTreutlein:
@benhylau Okay I think I fixed all the issues take a look at it again.

benhylau:
Thanks @rTreutlein try sha256?

rTreutlein:
@benhylau Yea tried it just messed up somewhere and lost some changes. Seems i forgot to readd that one.

benhylau:
Cool I will sort this over the wk, thanks!

scripts/hostapd/dnsmasq.conf
```diff
@@ -1,2 +1,5 @@
 interface=wlan0
+listen-address=10.0.0.1
+bind-interfaces
+server=208.67.222.222
```
benhylau: @rTreutlein Is it necessary to hardcode OpenDNS in here? This is what fixed the issue with your client waiting for an IP forever? Or was it the previous two lines?

benhylau:
@rTreutlein Sorry I have been swamped with a few other things, but am getting to this now :) I want to clarify the OpenDNS hardcode in dnsmasq.conf, and I forked this to #69
 to merge back updates from develop, and I will also move some files around so PSK and EAP are more separated into their own directories.

rTreutlein:
@benhylau It's not necessary to hard code the DNS server. I removed that line and it still works.

benhylau:
@rTreutlein I removed all three additions in dnsmasq.conf and everything still works. Typically those should not be needed. I have made all my changes on the #69 branch, if that looks okay to you I'll merge it. Closing this one, let's discuss on #69.

PR #65, opened 2017-03-03
Add example for adding ssh file

dcwalk:
Came up at virtual mesh that this step is confusing unless you've done it before. I agree!

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,11 @@ The following instructions will help you set up an encrypted mesh network on Ras
 
 1. Flash the SD card with [Raspbian Jessie Lite](https://www.raspberrypi.org/downloads/raspbian/).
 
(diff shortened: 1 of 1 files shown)


benhylau:
Not from Windows though right?

dcwalk:
Yeah, but i figured that would at least provide some context similar to the other examples?
`wget` doesn't work on Windows without an install. And we now have [bash on windows 10](http://www.windowscentral.com/how-install-bash-shell-command-line-windows-10)

benhylau:
Yes but all the commands like `wget` are run on the Pi. The `touch ssh` is not.

dcwalk:
Your call! I'm fine to leave, I thought just demonstrating something might be helpful? Could specify from linux?

PR #69, opened 2017-04-03
Add WPA-EAP support

benhylau:
See: https://github.com/tomeshnet/prototype-cjdns-pi/pull/63 by @rTreutlein

This branch can be tested with:

```
$ wget https://raw.githubusercontent.com/tomeshnet/prototype-cjdns-pi/eap/scripts/install && chmod +x install && TAG_PROTOTYPE_CJDNS_PI=eap ./install
```


rTreutlein:
Yea possibly the issues that lead me to add stuff to dnsmasq.conf where caused by something else (unclean install would be possible) but I can't quite remember as it was some time ago. If it works without them then just leave it like that.

benhylau:
@rTreutlein Thanks again for contributing!

PR #75, opened 2017-06-19
Automatically install NodeJS on ArmV6

darkdrgn2k:
**ISSUE**

When you run the install script on Pi1 install is blocked with following message

`## You appear to be running on ARMv6 hardware. Unfortunately this is not currently supported by the NodeSource Linux distributions. Please use the 'linux-armv6l' binary tarballs available directly from nodejs.org for Node.js v4 and later.
`

**RESOLUTION**

Modifications detect the armv6 processor and if found manually downloads node-v6.11.0-linux-armv6l.tar.gz  from nodejs.org, extracts/installs it then creates a link for binary from node to nodejs .

This allows for the installation to proceed without interruption.

### DIFF
--- a/scripts/install2
+++ b/scripts/install2
@@ -85,6 +85,7 @@ if ! [ "$(which nodejs)" ]; then
 pwd
         wget -O /tmp/node-v6.11.0-linux-armv6l.tar.gz https://nodejs.org/dist/v6.11.0/node-v6.11.0-linux-armv6l.tar.gz
         sudo tar xvfz /tmp/node-v6.11.0-linux-armv6l.tar.gz --strip 1 -C /
+	rm -rf /tmp/node-v6.11.0-linux-armv6l.tar.gz
         ln -s /bin/node /bin/nodejs
     else
        curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
--- a/scripts/install2
+++ b/scripts/install2
@@ -84,9 +84,9 @@ if ! [ "$(which nodejs)" ]; then
     if $(uname -m | grep -Eq ^armv6); then
 pwd
         wget -O /tmp/node-v6.11.0-linux-armv6l.tar.gz https://nodejs.org/dist/v6.11.0/node-v6.11.0-linux-armv6l.tar.gz
-        sudo tar xvfz /tmp/node-v6.11.0-linux-armv6l.tar.gz --strip 1 -C /
+        sudo tar xfz /tmp/node-v6.11.0-linux-armv6l.tar.gz --strip 1 -C /
 	rm -rf /tmp/node-v6.11.0-linux-armv6l.tar.gz
-        ln -s /bin/node /bin/nodejs
+        sudo ln -s /bin/node /bin/nodejs
     else
        curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
        sudo apt-get install nodejs -y
--- a/scripts/install2
+++ b/scripts/install2
@@ -79,10 +79,18 @@ fi
 
 # Get tools
 if ! [ "$(which nodejs)" ]; then
-    curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
-    sudo apt-get install nodejs -y
-fi
 
+#check for armv6 and install it manually instead
+    if $(uname -m | grep -Eq ^armv6); then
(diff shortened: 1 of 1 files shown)


scripts/install2
```diff
@@ -79,10 +79,20 @@ fi
 
 # Get tools
 if ! [ "$(which nodejs)" ]; then
-    curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
-    sudo apt-get install nodejs -y
-fi
 
+#check for armv6 and install it manually instead
+    if $(uname -m | grep -Eq ^armv6); then
+pwd
```
darkdrgn2k: Oops, this line should not be there, was used for debugging!

scripts/install2
```diff
@@ -79,10 +79,20 @@ fi
...
+        wget -O /tmp/node-v6.11.0-linux-armv6l.tar.gz https://nodejs.org/dist/v6.11.0/node-v6.11.0-linux-armv6l.tar.gz
+        sudo tar xfz /tmp/node-v6.11.0-linux-armv6l.tar.gz --strip 1 -C /
+	rm -rf /tmp/node-v6.11.0-linux-armv6l.tar.gz
```
benhylau: You have a mix of tabs and spaces, can you fix the formatting?

scripts/install2
```diff
@@ -79,10 +79,20 @@ fi
 
 # Get tools
 if ! [ "$(which nodejs)" ]; then
-    curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
-    sudo apt-get install nodejs -y
-fi
 
+#check for armv6 and install it manually instead
```
benhylau: Consistent with other comments please:

`# Check ...`

And replace "it" with "nodejs"?

scripts/install2
```diff
@@ -79,10 +79,20 @@ fi
...
+    if $(uname -m | grep -Eq ^armv6); then
+pwd
+        wget -O /tmp/node-v6.11.0-linux-armv6l.tar.gz https://nodejs.org/dist/v6.11.0/node-v6.11.0-linux-armv6l.tar.gz
+        sudo tar xfz /tmp/node-v6.11.0-linux-armv6l.tar.gz --strip 1 -C /
+	rm -rf /tmp/node-v6.11.0-linux-armv6l.tar.gz
+        sudo ln -s /bin/node /bin/nodejs
```
benhylau: I don't think this is needed right?

scripts/install2
```diff
@@ -79,10 +79,20 @@ fi
...
 if ! [ "$(which nodejs)" ]; then
-    curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
-    sudo apt-get install nodejs -y
-fi
 
+#check for armv6 and install it manually instead
+    if $(uname -m | grep -Eq ^armv6); then
+pwd
+        wget -O /tmp/node-v6.11.0-linux-armv6l.tar.gz https://nodejs.org/dist/v6.11.0/node-v6.11.0-linux-armv6l.tar.gz
+        sudo tar xfz /tmp/node-v6.11.0-linux-armv6l.tar.gz --strip 1 -C /
```
benhylau: I don't think this needs `sudo`

scripts/install2
```diff
@@ -79,10 +79,20 @@ fi
...
+	rm -rf /tmp/node-v6.11.0-linux-armv6l.tar.gz
+        sudo ln -s /bin/node /bin/nodejs
+    else
+       curl -sL https://deb.nodesource.com/setup_7.x | sudo -E bash -
+       sudo apt-get install nodejs -y
```
benhylau: Does this do the `node` -> `nodejs` symlink?

darkdrgn2k: yes, because nodejs is missing from the tar but your script requires nodejs to exist.

scripts/install2
darkdrgn2k: it is needed beucase the install script looks for nodejs

`if ! [ "$(which nodejs)" ]; then`

scripts/install2
darkdrgn2k: Actually you do need it because it extracts to / where you dont have perm to write to like /usr

scripts/install2
benhylau: Ahh I see, the apt-get installs as `nodejs`.

PR #77, opened 2017-06-23
Stronger checks and Balances

darkdrgn2k:
Status script updated to verify if hostpad is actually running or just installed, fixes detection of WIFI AP being active when it was not, just hostapd installed

IPFS Installation makes sure that pi user exists and that location and permissions are correct if running install outside of pi users.   Otherwise IPFS installation fails


scripts/status
```diff
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 INACTIVE='\e[1;31m[INACTIVE]\e[0m'
+STOPPED='\e[1;31m[STOPPED]\e[0m'
```
benhylau: Why is this not the same state as `INACTIVE`?

scripts/status
```diff
@@ -24,7 +25,11 @@ if [ "$(which mesh)" ]; then
 fi
 if [ "$(which hostapd)" ]; then
     if [ $(systemctl status hostapd.service | grep 'Active: ' | awk '{ print $2 }') = 'active' ]; then
-        echo -e "WiFi Access Point ............ $ACTIVE"
+        if [ -Z $(ps aux | grep hostapd | grep -v grep)  ]; then
```
benhylau: Should we just add this to the check, and only have `ACTIVE` vs. `INACTIVE`?

scripts/status
```diff
@@ -24,7 +25,11 @@ if [ "$(which mesh)" ]; then
...
+            echo -e "WiFi Access Point ........... $STOPPED"
```
benhylau: I think 2 less `..` to keep alignment

scripts/ipfs/install
```diff
@@ -5,7 +5,10 @@ set -e
 GO_IPFS_VERSION=v0.4.6
 
 BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-
+# Check for PI user if missing add it
+if [ -z $(cat /etc/passwd | grep pi:x) ];then
+    useradd pi
```
benhylau: Can't we just use `~` to get the existing home directory instead of making a new user? This IPFS installer wouldn't be the appropriate place to make a new user anyway.

darkdrgn2k: I looked at that but some of the configs that get copied over are hardcoded to /home/pi/.ipfs  
This is why you need the pi user.  Perhaps a revisit of this is in order but i figured this was a quick patch.

scripts/status
darkdrgn2k: Seems aligned to me ?!?!
Stopped needed one less when i moved it to Inactive

benhylau:
What is the base image you used for this? Something from here? https://www.armbian.com/orange-pi-zero/

darkdrgn2k:
i used the nightly build of 4.2 kernel from armbian 

its not very evident with the new site where to find it but basically
* https://www.armbian.com/ 
*Select your Board
* other download options and archive
*Nightly
* pick your poison

in my case it was 
https://dl.armbian.com/orangepione/nightly/
and
https://dl.armbian.com/orangepilite/nightly/

benhylau:
I also have trouble finding things with the new site. I will just link the nightly downloads.

scripts/status
benhylau: Changed to:
```
if [ "$(which hostapd)" ]; then
    if [ $(systemctl status hostapd.service | grep 'Active: ' | awk '{ print $2 }') = 'active' ] && [ -n "$(ps aux | grep hostapd | grep -v grep)" ]; then
        echo -e "WiFi Access Point .......... $ACTIVE"
    else
        echo -e "WiFi Access Point .......... $INACTIVE"
    fi
fi
```

PR #78, opened 2017-06-23
Orange pi

darkdrgn2k:
Added support for Orange PI Lite and Orange Pi One (and possibly others?)
* Added detection for Allwinner CPU 
* Disabled predictable network interface naming to keep from changing and making mesh script to fail

### DIFF
--- a/scripts/install2
+++ b/scripts/install2
@@ -6,26 +6,30 @@ TAG_CJDNS=cjdns-v19.1
 
 # Get Raspberry Pi version and set flags accordingly
 RPI_REVISION=`sed -rn 's/Revision\s+\:\s+([0-9a-z_\-\s\,\(\)]+)/\1/p' /proc/cpuinfo`
+BOARD_FAMILY='Raspberry PI'
+
 CJDNS_BUILD_CMD="sudo Seccomp_NO=1 NO_NEON=1 ./do"
 
-if [[ $RPI_REVISION == *"900092"* || $RPI_REVISION == *"900093"* ]]; then
-    echo -e "\e[1;32mStarting installation on Raspberry Pi Zero...\e[0m"
-    CJDNS_BUILD_CMD="sudo Seccomp_NO=1 NO_NEON=1 CFLAGS=\"-s -static -Wall\" ./do"
-elif [[ `cat /proc/cpuinfo  | grep Hardware | awk '{print $3}'` == 'Allwinner' ]]; then
+if [[ `cat /proc/cpuinfo  | grep Hardware | awk '{print $3}'` == 'Allwinner' ]]; then
+    BOARD_FAMILY='Orange PI'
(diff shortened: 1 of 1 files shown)


scripts/install2
```diff
@@ -11,6 +11,10 @@ CJDNS_BUILD_CMD="sudo Seccomp_NO=1 NO_NEON=1 ./do"
 if [[ $RPI_REVISION == *"900092"* || $RPI_REVISION == *"900093"* ]]; then
     echo -e "\e[1;32mStarting installation on Raspberry Pi Zero...\e[0m"
     CJDNS_BUILD_CMD="sudo Seccomp_NO=1 NO_NEON=1 CFLAGS=\"-s -static -Wall\" ./do"
+elif [[ `cat /proc/cpuinfo  | grep Hardware | awk '{print $3}'` == 'Allwinner' ]]; then
+    echo -e "\e[1;32mStarting installation on Orange Pi ???... \e[1;31m(EXPERIMENTAL) \e[0m"
+    CJDNS_BUILD_CMD="sudo Seccomp_NO=1 CFLAGS=\"-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm\" ./do"
+    RPI_REVISION="OPI"
```
benhylau: `OPI` is not a `RPI_REVISION`. You need to at the top of the file scan for whether this is a Raspberry Pi board or Orange Pi board and have a `BOARD_FAMILY` or similarly named variable, then go into revision detect login after matching per board family. It also needs stronger checks than just `Allwinner`.

scripts/install2
```diff
@@ -77,6 +81,10 @@ if [ "${#MESH_NAME}" == 0 ]; then
     export MESH_NAME="tomesh"
 fi
 
+if [[ $RPI_REVISION == "OPI" ]]; then
```
benhylau: All the OPI-specific stuff should go into one single blob. This will quickly become unmanageable with `$RPI_REVISION == X` scattered all over the files.

scripts/install2
darkdrgn2k: I tried not to make it as intrusive by introducing a second variable, but i agreea that board_family is a better fix.

I was speaking with the ARMBAIN people about identifying boards and they noted that it is impossible to identify  a Orange PI live vs Orange PI One board that is why they have different images for each.

I'm open to suggestions on how to detect it, or does it need to be stronger at this point in time.

scripts/install2
darkdrgn2k: I Agree, however in this case I think it is smart to separate DETECTION and ACTION. 

One reason is i looked at the flow of the script and no changes are made in the first part of the script. Script currently is split into  detect the revision, ask a few question and THEN make changes

This is why i opted out to start making changes AFTER the last question is asked.

PR #82, opened 2017-07-06
Added function to detect meshable aps

darkdrgn2k:
Added false positive on Orange Pi Zeron onboard wifi
Better way of detecting iw interface when not named wlan0 (ie 2nd run)


darkdrgn2k:
Baa sorry again.

PR #86, opened 2017-09-04
Merge develop into master

benhylau:


darkdrgn2k:
Only thing I found is the killall doesn't solve the problem. But it may have been an anomoly that I  noticed from a crashed package install..


I'll keep an eye on it. It's not showstopper and can be fixed later as if it happens it happens rearly and is not a direct bug to the script but rather environment. 

As long as it works on the pi 3 (don't have one can't test) I think it's good to go

scripts/install2
```diff
@@ -42,10 +42,12 @@ fi
 
 echo -e "\e[1;32mStarting installation on ${BOARD_FAMILY} ${BOARD_NAME} (${BOARD_REVISION})...\e[0m"
 
+
+# Disable predictable network interface naming on Orange Pi
```
benhylau: Now it's not just for Orange Pi right?

benhylau: I mean, the comment doesn't need to say `Orange Pi` anymore :)

PR #87, opened 2017-09-15
Hyperboria Dns server

darkdrgn2k:
Update network config file on boot to replace dns servers with hyperboria capable ones.
DHCP servers will still be added as per normal


darkdrgn2k:
Merging common feature set into one PR

PR #89, opened 2017-09-17
Mesh config update

darkdrgn2k:
Using UDev to configure mesh point when device is plugged in instead of hard coded script.

port from Orange Mesh


darkdrgn2k:
New PR Coming

PR #91, opened 2017-09-18
[WIP] New configuration for hostap

darkdrgn2k:
new config for HOSTAP
using systemd-network just like the AP pull

limited MESH POINT and AP by driver


darkdrgn2k:
New PR Coming

PR #96, opened 2017-09-24
Orange Pi Detection

darkdrgn2k:
Better detection for Orange Pi
New compile flags for Orange Pi One (pulling up to 40mpbs now)


scripts/install2
```diff
@@ -14,12 +14,16 @@ BOARD_HARDWARE=$(cat /proc/cpuinfo  | grep Hardware | awk '{print $3}' | head -n
 # Possibly use /sys/firmware/devicetree/base/model
 if [[ "$BOARD_HARDWARE" == 'Allwinner' || "$BOARD_HARDWARE" == 'Generic' ]]; then    
     BOARD_FAMILY="Orange Pi"
-    BOARD_NAME="Generic"
+    source /etc/armbian-image-release
```
benhylau: Do this after all the variable assignments

scripts/install2
```diff
@@ -14,12 +14,16 @@ BOARD_HARDWARE=$(cat /proc/cpuinfo  | grep Hardware | awk '{print $3}' | head -n
...
     BOARD_REVISION="experimental"
-    CJDNS_BUILD_CMD="sudo Seccomp_NO=1 CFLAGS=\"-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm\" ./do"
+    CJDNS_BUILD_CMD="sudo Seccomp_NO=1 CFLAGS=\"-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm\" ./do"	
+    if [[ '$BOARD_NAME' == 'One' ]]; then
+        CJDNS_BUILD_CMD="sudo Seccomp_NO=1 CFLAGS=\"-s -static -Wall -march=armv7-a -mtune=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard\" ./do"
+    fi
 elif [[ "$BOARD_HARDWARE" == 'sun50iw1p1' ]]; then
     BOARD_FAMILY="Orange Pi"
-    BOARD_NAME="H5"
+    source /etc/armbian-image-release
```
benhylau: Same as above

scripts/install2
```diff
@@ -14,12 +14,16 @@ BOARD_HARDWARE=$(cat /proc/cpuinfo  | grep Hardware | awk '{print $3}' | head -n
 # Possibly use /sys/firmware/devicetree/base/model
 if [[ "$BOARD_HARDWARE" == 'Allwinner' || "$BOARD_HARDWARE" == 'Generic' ]]; then    
     BOARD_FAMILY="Orange Pi"
-    BOARD_NAME="Generic"
+    source /etc/armbian-image-release
+    BOARD_NAME=`echo $BOARD_NAME | sed 's/Orange Pi //g'`
```
benhylau: Isn't `BOARD_NAME` always "Generic" here?

scripts/install2
```diff
@@ -14,12 +14,16 @@ BOARD_HARDWARE=$(cat /proc/cpuinfo  | grep Hardware | awk '{print $3}' | head -n
...
     BOARD_REVISION="experimental"
-    CJDNS_BUILD_CMD="sudo Seccomp_NO=1 CFLAGS=\"-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm\" ./do"
+    CJDNS_BUILD_CMD="sudo Seccomp_NO=1 CFLAGS=\"-s -static -Wall -mfpu=neon -mcpu=cortex-a7 -mtune=cortex-a7 -fomit-frame-pointer -marm\" ./do"
```
benhylau: This should be in the `else`.

scripts/install2
```diff
@@ -42,12 +46,7 @@ fi
 
 echo -e "\e[1;32mStarting installation on ${BOARD_FAMILY} ${BOARD_NAME} (${BOARD_REVISION})...\e[0m"
 
-<<<<<<< HEAD
-
-# Disable predictable network interface naming newer kernels
```
benhylau: ?

scripts/install2
darkdrgn2k: it gets sets with this line 
source /etc/armbian-image-release

scripts/install2
darkdrgn2k: Merge errors I thought I committed the fix. Keeps creeping up from. Somewhere :#

scripts/install2
benhylau: `/etc/armbian-image-release` is a file that comes with the Armbian image, and we are relying on it to set our `BOARD_NAME` variable?

darkdrgn2k: Correct
From my research its the ONLY way to identify the board 100% is by the image that is flashed. Each image includes this file

its content is as follows (for an zero)

root@orangepizero:~# cat /etc/armbian-image-release 
# PLEASE DO NOT EDIT THIS FILE
BOARD=orangepizero
BOARD_NAME="Orange Pi Zero"
VERSION=5.32.170920
LINUXFAMILY=sun8i
BRANCH=dev
ARCH=arm
IMAGE_TYPE=nightly
BOARD_TYPE=conf
INITRD_ARCH=arm
KERNEL_IMAGE_TYPE=zImage

benhylau: I think it's better to grep for the "BOARD_NAME" string value and set the local `BOARD_NAME` variable to the "Zero" than to source the entire file.

darkdrgn2k: It seems allot more cleaner/simpler then

``` BOARD_NAME=$(cat /etc/armbian-image-release | grep BOARD_NAME | awk -F \= '{print $2}' | tr -d \")```

but i can go either way...

scripts/install2
darkdrgn2k: oh it was a bad commit, its already been corrected from a previous commit somewhere..

scripts/install2
benhylau: Don't need the `cat`, just `grep`. But this way it's more readable and we won't accidentally run unintended things in `/etc/armbian-image-release`.

PR #99, opened 2017-10-02
Dialog gui for install options

benhylau:


scripts/installGUI
```diff
@@ -0,0 +1,28 @@
...
+export WITH_EXTRA_TOOLS=false
+
+sudo apt-get install dialog
+dialog --checklist "Install Options:" 20 70 10 \
+        "WITH_MESH_POINT" "Mesh Point" on \
+        "WITH_WIFI_AP" "Access Point" on \
+        "WITH_IPFS" "IPFS Service" on \
+        "WITH_PROMETHEUS_NODE_EXPORTER" "Node Exporter" on \
+        "WITH_PROMETHEUS_SERVER" "Prometheus" on \
+        "WITH_GRAFANA" "Grafana" on \
```
benhylau: - [ ] Add `WITH_H_DNS` `WITH_H_NTP` `WITH_FAKE_HWCLOCK`.
- [ ]  Default everything to off.
- [ ] Rename file to `install-gui`

benhylau:
@darkdrgn2k So basically we run this instead right?
```
$ wget https://raw.githubusercontent.com/tomeshnet/prototype-cjdns-pi/master/scripts/install-gui && chmod +x install-gui && ./install-gui
```

darkdrgn2k:
Patched it so its

`$ wget https://raw.githubusercontent.com/tomeshnet/prototype-cjdns-pi/master/scripts/install && chmod +x install && ./install gui
`

But I think you can shrink it to

`$ wget https://raw.githubusercontent.com/tomeshnet/prototype-cjdns-pi/master/scripts/install && . ./install gui
`

scripts/installGUI
```diff
@@ -0,0 +1,93 @@
...
+    export SET_PROMETHEUS_SERVER=on
+    export SET_GRAFANA=on
+    export SET_EXTRA_TOOLS=on
+    export SET_H_DNS=on
+    export SET_H_NTP=on
+    export SET_FAKE_HWCLOCK=on
+fi
+
+dialog --backtitle "TOMesh Prototype Node" --title "Select Features" --checklist "Features:" 20 70 10 \
+        "WITH_MESH_POINT" "Mesh Point" $SET_MESH_POINT \
```
garrying: Thinking the first column should be the title case label instead of the `export` name. e.g.
`"Mesh Point" "Allow peering through an 802.11s interface" $SET_MESH_POINT \`

scripts/installGUI
```diff
@@ -0,0 +1,93 @@
...
+export SET_H_NTP=off
+export SET_FAKE_HWCLOCK=off
+
+sudo apt-get install dialog
+
+dialog --backtitle "TOMesh Prototype Node" --title "Select feature starting point (Customize next)" --menu "Presets" 20 70 10 \
+        BASIC "Basic" \
+        HYPE  "Basic With Hyperboria" \
+        FULL  "Install Everything" \
+        CANCEL "Cancel" \
```
garrying: This option seems redundant given that checklist has a `cancel` option, which isn't working as fast as I can tell—takes you to the next screen.

scripts/installGUI
```diff
@@ -0,0 +1,93 @@
...
+export SET_GRAFANA=off
+export SET_EXTRA_TOOLS=off
+export SET_H_DNS=off
+export SET_H_NTP=off
+export SET_FAKE_HWCLOCK=off
+
+sudo apt-get install dialog
+
+dialog --backtitle "TOMesh Prototype Node" --title "Select feature starting point (Customize next)" --menu "Presets" 20 70 10 \
+        BASIC "Basic" \
```
garrying: Are we generally using `Basic` to describe the Mesh Point w/ AP? Seems ambiguous in this context.

scripts/installGUI
```diff
@@ -0,0 +1,93 @@
...
+if [ "$res" == "" ]; then
+    exit 0
+fi
+rm -rf /tmp/answer
+
+for item in $res;do
+   t="export $item=true"
+   eval $t
+done
+./install2
```
garrying: Are we able to export or bring in the `mesh network` name prompt? Seems a bit disjointed to have it be presented after confirming features.

scripts/installGUI
darkdrgn2k: Action is not by choice

The first column is what the dialog app sets the value the second column is the description.

scripts/installGUI
darkdrgn2k: This is doable using something like 
 
`dialog --inputbox "Enter your name:" 8 40 `

But i think this that at the moment the install script has an export for this

scripts/installGUI
darkdrgn2k: Hmm didnt notice that.. 

We can remove that and just replace line 37 with
`if [ "$res" == "" ]; then`

scripts/installGUI
darkdrgn2k: This was just a quick slap together list as a possible idea to extend @benhylau of setting  everything off, this way there are PRESETS.

We can create whatever groups we want i guess that needs to be decided. Its just a starting point:)

AP is a weird thing because in current state it ONLY works with the onboard wifi on a rpi3 and doesn't actually ask you if you want to install if unless your running an RPI3

MeshScripts PR will fix this

scripts/installGUI
darkdrgn2k: The other issue is that some of the export are really long, like the prom node exporter, if we can make it smaller we can allow for more room for the friendly names.

scripts/installGUI
darkdrgn2k: We should also do this for AP PSK/EAP 

`dialog --backtitle "Select WPA" \
  --radiolist "WPA Type" 10 40 4 \
        1 PSK on \
        2 EAP off 
`
and Password using the inputbox above.

darkdrgn2k:
Added WITH_MESH_ID to set the meshpoint SSID name
Added prompt to check Mesh point ID in gui

README.md
```diff
@@ -56,6 +56,7 @@ Many models of Orange Pi hardware running [Armbian](https://www.armbian.com/) ar
 | `WITH_H_NTP`                    | None                                           | Set to `true` if you want to use a Hyperboria-compatible NTP server: `fc4d:c8e5:9efe:9ac2:8e72:fcf7:6ce8:39dc` |
 | `WITH_FAKE_HWCLOCK`             | None                                           | Set to `true` if you want to force hwclock to store its time every 5 minutes. |
 | `WITH_EXTRA_TOOLS`              | None                                           | Set to `true` if you want to install non-essential tools useful for network analysis: vim socat oping bmon iperf3 |
+| `WITH_MESH_ID`                  | None                                           | Set the mesh ID for use with `WITH_MESH_POINT` |
```
benhylau: These are all boolean flags to enable/disable features. I don't want to start piling other configs to this list. Can we get the mesh ID some other way? How do we get the WPA2 password?

scripts/install
```diff
@@ -12,17 +12,17 @@ if [ -z "$TAG_PROTOTYPE_CJDNS_PI" ]; then
     TAG_PROTOTYPE_CJDNS_PI=master
 fi
 
-if ! [ -z "`ps xa | grep dpkg | grep -v grep`" ]; then 
+if ! [ -z "`ps xa | grep dpkg | grep -v grep`" ]; then
```
benhylau: Merge back develop and resolve the conflict there.

scripts/install
```diff
@@ -46,6 +46,12 @@ export WITH_PROMETHEUS_NODE_EXPORTER
 export WITH_PROMETHEUS_SERVER
 export WITH_GRAFANA
 export WITH_EXTRA_TOOLS
+export WITH_MESH_ID
 
 # Run the actual installation script
-./install2
+
+if [ "$1" = "gui" ]; then
```
benhylau: Need README instructions on how to use GUI

scripts/install2
```diff
@@ -188,12 +188,20 @@ if [ -z "$WITH_EXTRA_TOOLS" -o "$WITH_EXTRA_TOOLS" != "true" -a "$WITH_EXTRA_TOO
     fi
 fi
 
-# Prompt for name of the mesh network
-read -p "Enter the name of your mesh network (default: tomesh): " -r
-export MESH_NAME=`echo $REPLY | sed 's/ //g'`
-if [ "${#MESH_NAME}" == 0 ]; then
-    export MESH_NAME="tomesh"
+#Only if mesh point needs to be set
```
benhylau: Format

scripts/installGUI
```diff
@@ -0,0 +1,102 @@
...
+
+if [ "$res" == "HYPE" ]; then
+    export SET_MESH_POINT=on
+    export SET_WIFI_AP=on
+    export SET_PROMETHEUS_NODE_EXPORTER=on
+    export SET_EXTRA_TOOLS=on
+    export SET_H_DNS=on
+    export SET_H_NTP=on
+    export SET_FAKE_HWCLOCK=on
+
```
benhylau: Format

scripts/installGUI
```diff
@@ -0,0 +1,102 @@
...
+fi
+rm -rf /tmp/answer
+
+for item in $res;do
+   t="export $item=true"
+   eval $t
+done
+
+if [ "$WITH_MESH_POINT" == "true" ]; then
+	dialog --backtitle "$dialogTitle" $additionalParam  --title "Meshing Setup" --inputbox "802.11s Mesh Name"  10 70  "tomseh" 2> /tmp/answer
```
benhylau: Tabs

benhylau:
This needs more discussion, de-scoping from next release since it is not required for Stretch support.

benhylau:
Closing this PR since we never came to a good solution for https://github.com/tomeshnet/prototype-cjdns-pi/pull/99/files#r147587481

Re-open in the future if this feature is still needed.

PR #102, opened 2017-10-20
Patch for interface on stretch

darkdrgn2k:
Patch for interface error on stretch 

I dont have a PI3 so can you please test it, in theory it should work

### DIFF
--- a/scripts/hostapd/install
+++ b/scripts/hostapd/install
@@ -2,6 +2,7 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


benhylau:
I think we can check for non-empty $START instead:

```
if [ ! -z $START ]; then
    END=$(expr $START + 3)
    sudo sed -i "${START},${END}d" /etc/network/interfaces
fi
```

PR #107, opened 2017-10-24
Cjdns performance patch

darkdrgn2k:
Increases performance by 25%


darkdrgn2k:
added benhylau's cherry picked method to apply upstream patch rather then simply disabling it

PR #108, opened 2017-10-24
Install Process for cjdns

darkdrgn2k:
Currently the build process ignores the newly build cjdns if /usr/bin/cjdroute exists.

This is very confusing if you installed it previously and your banging your head against the wall wondering why the specs are not better with the new patches :)

Patch will remove old cjdns in favor of the newly compiled one

### DIFF
--- a/scripts/install2
+++ b/scripts/install2
@@ -226,13 +226,15 @@ fi
 if ! [ -x "/opt/cjdns/cjdroute" ]; then
     here=`pwd`
     cd /opt/cjdns && eval $CJDNS_BUILD_CMD && cd $here
+    if ! [ -x "/usr/bin/cjdroute" ]; then
+        sudo rm -f /usr/bin/cjdroute
+    fi
 fi
 
 # Install cjdns to /usr/bin
 if ! [ -x "/usr/bin/cjdroute" ]; then
-    rm -rf /usr/bin/cjdroute
+    sudo cp /opt/cjdns/cjdroute /usr/bin/cjdroute
 fi
-sudo cp /opt/cjdns/cjdroute /usr/bin/cjdroute
 
 # Generate cjdns configurations
 if ! [ -f "/etc/cjdroute.conf" ]; then
--- a/scripts/install2
+++ b/scripts/install2
@@ -230,8 +230,9 @@ fi
 
 # Install cjdns to /usr/bin
 if ! [ -x "/usr/bin/cjdroute" ]; then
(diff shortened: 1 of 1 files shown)


scripts/install2
```diff
@@ -230,8 +230,9 @@ fi
 
 # Install cjdns to /usr/bin
 if ! [ -x "/usr/bin/cjdroute" ]; then
-    sudo cp /opt/cjdns/cjdroute /usr/bin/cjdroute
+    rm -rf /usr/bin/cjdroute
```
benhylau: Don't you need sudo here too?

benhylau:
You are also making the assumption that nothing else changed in your new version except for `cjdroute`: https://github.com/tomeshnet/prototype-cjdns-pi/blob/master/scripts/uninstall#L7-L13

This only addresses a narrow use case of prototyping patches. If `install2` is to support upgrade, it will need more than just swapping the cjdroute binary.

darkdrgn2k:
In short: Why spend the time compiling if your not even going to apply it!

darkdrgn2k:
Looking at the code again its in the wrong spot, moved it into the compile block, so that way if it compiles and /usr/bin/cjdroute exists, only then will it delete it.

that way the compile is not wasted!

darkdrgn2k:
Sudo was added to the wrong line, its corrected now.

scripts/install2
```diff
@@ -226,6 +226,9 @@ fi
 if ! [ -x "/opt/cjdns/cjdroute" ]; then
     here=`pwd`
     cd /opt/cjdns && eval $CJDNS_BUILD_CMD && cd $here
+    if ! [ -x "/usr/bin/cjdroute" ]; then
+        sudo rm -f /usr/bin/cjdroute
```
benhylau: Can we close this PR and make a single commit that puts `sudo rm -f /usr/bin/cjdroute` before `sudo cp /opt/cjdns/cjdroute /usr/bin/cjdroute`, like this:

```
# Install cjdns to /usr/bin
if ! [ -x "/usr/bin/cjdroute" ]; then
    sudo rm -f /usr/bin/cjdroute
    sudo cp /opt/cjdns/cjdroute /usr/bin/cjdroute
fi
```

Because the removal is part of install and not build. I think this is most clear. And new commit as all the history here are a little confusing.

benhylau:
Addressed by https://github.com/tomeshnet/prototype-cjdns-pi/pull/110

PR #118, opened 2017-11-24
DNSMasq and Hyeprborea DNS

darkdrgn2k:
Tell DNSMasq to use new hyperborea dns servers when dns-h enabled.


scripts/hostapd/install
```diff
@@ -83,9 +83,14 @@ fi
 if [ -f /etc/dnsmasq.conf ]; then
     sudo cp /etc/dnsmasq.conf /etc/dnsmasq.conf.bak
 fi
+
 sudo cp "$BASE_DIR/dnsmasq.conf" /etc/dnsmasq.conf
+if [ ! -z "$WITH_H_DNS" -a "$WITH_H_DNS" == "true" ]; then
+   echo "server=$H_DNS_SERVER_0" | sudo tee --append /etc/dnsmasq.conf
```
benhylau: This requires `h-dns` module to be run before `hostapd`. We don't currently enforce ordering of module installation. We also won't be able to run `hostapd` install on its own after initial installation with this change.

`WITH_*` flags should only be used to switch whether a module's install will run or not.

darkdrgn2k: I know what you mean

Assumption - Install from scratch
Unless you want to hard code the H-dns servers into the hostapd script and have them in two differnt spots i dont know how else to fix this :(

benhylau: Hard-code for now and transition to specifying DNS in config file later

darkdrgn2k: couldn't get rid of the h_dns check, because if the node is not a hyperborea node it will break DNS resolution on AP. but otherwise it break dns resolution if its just a hyperbola node... so yeh... its a bit of a kludge but we can fix it later :) it will not "work" only if the two items are installed separately or in the wrong order. But it will work more of the time then it does now.

PR #124, opened 2017-12-20
Configure Internet gateway using cjdns iptunnel

benhylau:


scripts/hostapd/install
```diff
@@ -27,10 +27,10 @@ echo "iface lo inet loopback" | sudo tee --append /etc/network/interfaces > /dev
 echo "" | sudo tee --append /etc/network/interfaces > /dev/null
 echo "allow-hotplug wlan0" | sudo tee --append /etc/network/interfaces > /dev/null
 echo "iface wlan0 inet static" | sudo tee --append /etc/network/interfaces > /dev/null
-echo "    address 10.0.0.1" | sudo tee --append /etc/network/interfaces > /dev/null
-echo "    netmask 255.255.255.0" | sudo tee --append /etc/network/interfaces > /dev/null
-echo "    network 10.0.0.0" | sudo tee --append /etc/network/interfaces > /dev/null
-echo "    broadcast 10.0.0.255" | sudo tee --append /etc/network/interfaces > /dev/null
+echo "    address 10.10.0.1" | sudo tee --append /etc/network/interfaces > /dev/null
```
darkdrgn2k: What is the reason for shifting to a new subnet?

scripts/hostapd/install
```diff
@@ -27,10 +27,10 @@ echo "iface lo inet loopback" | sudo tee --append /etc/network/interfaces > /dev
...
+echo "    netmask 255.255.255.128" | sudo tee --append /etc/network/interfaces > /dev/null
```
darkdrgn2k: Setting to such a non-standard net mask is not really needed. it does not have to match the dhcp scope we could just leave it as 255.255.255.0

scripts/hostapd/dnsmasq.conf
```diff
@@ -1,2 +1,2 @@
-dhcp-range=10.0.0.2,10.0.0.127,255.255.255.0,12h
+dhcp-range=10.0.0.2,10.0.0.126,12h
```
darkdrgn2k: Network has been changed to 10.0.0.1 so if you want to keep this it should be 
```dhcp-range=10.10.0.2,10.10.0.126,12h``` or the network reset to 10.0.0.1 below.

Comment (benhylau) [reply scripts/hostapd/dnsmasq.conf]: My bad, these should be `10.10.0.x`

File: scripts/hostapd/install (hunk 1 above)
Comment (benhylau) [reply scripts/hostapd/install]: So we can have `10.11.0.0/16` for VPN?

File: scripts/hostapd/install (hunk 2 above)
Comment (benhylau) [reply scripts/hostapd/install]: But we discussed this on chat, to use `255.255.255.128` to match DHCP scope, no? Have I misunderstood?

File: scripts/hostapd/install (hunk 1 above)
Comment (darkdrgn2k) [reply scripts/hostapd/install]: So first the config change IS correct and it WILL work

However i will point out that the change is NOT NEEDED and here is why:

The existing config  is already "10.0.0.0" netmask "255.255.255.0" (or 10.0.0.0/24)
This means the  range is 10.0.0.0 - 10.0.0.255

It does not overlap with 10.10.0.0/24 who's  range is 10.10.0.0-10.10.0.255
So you don't really need to make this change and still have VPN on 10.10.0.0

Why this whole thing happened:
If  the netmask was 255.0.0.0 for either ranges it would cause a problem because the range would be 10.0.0.0 - 10.255.255.255 and hence overlap each other

I think what happened is when you where testing it at first you assigned the ip as such
```ifconfig tun0 10.10.0.1```
instead of
```ifconfig tun0 10.10.0.1/24```

In the 10.x.x.x is defined as a CLASS A network, and as such it defaults to a netmaks of 255.0.0.0.  

This means that 10.0.0.0 and 10.10.0.0 overlapped and  caused problems

When we talked i didn't not realize what happend and did not have the config in front of me, assumed that 10.0.0.0 was set to 255.0.0.0 as well (it was not)

File: scripts/hostapd/install (hunk 2 above)
Comment (darkdrgn2k) [reply scripts/hostapd/install]: TL;DR - Range NEVER matches the netmask, its always a smaller subset of iwhat the netmask defines.

So again the config IS correct but not needed (kinda see below for a correction)

DHCP SCOPE does not need to "match" the available network addresses, in fact its usually a small subset of the available range, and never ALL of it (because the dhcp server needs an ip (ie 10.0.0.1))

The only limit is that the scope MUST BE contained within the range of the network

```dhcp-range=10.0.0.2,10.0.0.127,255.255.255.0,12h ```
This example asks dhcp to issue ip addresses between 2 to 127 with the netmask of 255.255.255.0. In such a network you can have ips all the way up to 254 but DHCP simply wont offer them up. This is usually reserved for things like STATIC IP ADDRESSES for a printer, or server or other.


SIDE NOTE
> Something i have seen with "public" wifi networks is that they dont make the scope big enough and the expiry time short enough. So in this example the first 127 people that join the network will work fine the 128 person even though he/she joined 9 hours after everyone else left, wont get on because there are no more ip addresses left to offer until the old ones expire. This is not a problem with out setup but in a roaming environment it may be.

Now Reading my  previous comment, I think there is an error in this line actually
```dhcp-range=10.0.0.2,10.0.0.126,12h ```

This would let the computer define its own subnet, and as we saw in the previous comment the subset would be assigned by default as 10.0.0.0/16

You did not notice this problem because although the subnets the overlap they don't do it on the router but rather  on your laptop (or other device). Because your laptop does not know of  10.10.0.0/24 (its not one of the interfaces there) it "doesnt matter"

The issue would arise if you tried to (for example) ping 10.10.0.1
your computer would think that 10.10.0.1 was on the SAME LAN as 10.0.0.0 because of the incorrect NETMASK of 255.0.0.0 and try to switch it instead of routing it and it would never get to where its going.

File: scripts/hostapd/install (hunk 1 above)
Comment (darkdrgn2k) [reply scripts/hostapd/install]: TL;DR - The ranges actually never overlap with the old config.
10.0.0.0/24  10.10.0.0/24

Comment (benhylau) [reply scripts/hostapd/install]: Ahhhhh yes I saw `10.0.0.0` and just assumed it was a `255.0.0.0`. Now I remember that because we know the Raspberry Pi cannot handle so many clients anyways, we made it a `255.255.255.0`, and DHCP assign only 128 of those (which still is excessive).

File: scripts/hostapd/install (hunk 2 above)
Comment (benhylau) [reply scripts/hostapd/install]: The laptop does pick up the subnet mask from **/etc/network/interfaces** even when I don't specify it on the mask in `dhcp-range` in **dnsmasq.conf**:

screen shot 2017-12-22 at 2 11 40 pm

But here I have `dhcp-range=10.10.0.2,10.10.0.126,12h` not `10.0`.

File: scripts/hostapd/dnsmasq.conf [hunk 4]
```diff
@@ -1,2 +1,2 @@
-dhcp-range=10.0.0.2,10.0.0.127,255.255.255.0,12h
+dhcp-range=10.0.0.2,10.0.0.254,4h
```
benhylau: I am DHCP-ing the full range (no room for static) and changing to 4h lease.

darkdrgn2k: ```dhcp-range=10.0.0.2,10.0.0.254,4h```

should be 

```dhcp-range=10.0.0.2,10.0.0.254,255.255.255.0,4h```

So there is no confusion on the assigned ip addresses that they are /24  on the client side

scripts/hostapd/install
darkdrgn2k: ok cool... looks like dnsmasq already pulls it from the interface and not the ancient "class a , b ,c " definition :+1:

benhylau: I will add in dnsmasq anyways so there is no ambiguity

scripts/hostapd/install
ForgivenNin: I know that this addition has been added, and that this may be a dumb question, but is there any specific type of VPN that this will be able to work with (like github.com/pivpn/pivpn or OpenVPN).

The reason I ask is that since cjdns is P2P, I assumed that a VPN would mess up some things...

darkdrgn2k: Kind of depends what your trying to do.

NON-CJDNS  traffic cannot exist on CJDNS 
CJDN's IPTUNNEL basically creates a VPN through CJDNS's network to allow non-cjdns traffic through. This is how you can "share" normal internet access with peers on CJDNS/Hyperborea

You can still run OpenVPN  or whatnot on your node.  The trick would be to to engineer your traffic to actually do something.

PR #93, opened 2017-09-19
[WIP] Mesh and HostAP Scripts (uDev/systemd)

darkdrgn2k:
Picked from Orange Mesh scripts to help identify hardware being plugged in and assign tasks to it accordingly

Many files from HOSTAP removed, and should be cherry picked back in as needed.

PR #126, opened 2017-12-21
[WIP] Config file

darkdrgn2k:
New configuration method for node using config file


benhylau:
So what is the status of this?

darkdrgn2k:
See #130 for status update

scripts/mesh-point/install
```diff
@@ -16,7 +16,7 @@ sudo cp "$BASE_DIR/mesh.wlan.remove" "/usr/local/sbin"
 sudo cp "$BASE_DIR/mesh.hostapd.setup" "/usr/local/sbin"
 
 sudo cp "$BASE_DIR/hostapd@.service" "/etc/systemd/system/hostapd@.service" 
-sudo cp "$BASE_DIR/50-hostapd.rules" /etc/udev/rules.d 
+sudo cp "$BASE_DIR/51-hostapd.rules" /etc/udev/rules.d
```
benhylau: @darkdrgn2k @hamishcoleman I am guessing the mesh-orange equivalent of this is:

`(sd_fat_fs_root)/conf.d/05_network_interfaces.tar` containing these files in the archive?

```
udev/rules.d/50-mesh.wlan.rules
udev/rules.d/50-mesh.wlan.remove.rules
udev/rules.d/51-hostapd.rules
```

How would the following be mapped? They ship with the base OS and are not overridable via mesh-orange **conf.d**?

```
/usr/local/sbin/mesh.wlan.setup
/usr/local/sbin/mesh.wlan.remove
/usr/local/sbin/mesh.hostapd.setup
```

cc. @Famicoman

darkdrgn2k: There, is no direct parallel to Mesh Orange (of mesh scripts which is a port of current Mesh Orange code into prototype)
in Mesh Orange currently *.link files contain the configuration of individual interfaces

The above mentioned files hold NO unique configurations (and therefore can be part of the BASE operating system). They simply act based on a config file.

The udev rules (below) trigger the scripts (lower down) when a Network Interface is changed (add/remove)

```
# When any network interface is added trigger /usr/local/sbin/mesh.wlan.setup 
# Trigger hostapd@int.service (which now has its own filter if it will actually run)
udev/rules.d/50-mesh.wlan.rules 

# When any network interface is Removed trigger /usr/local/sbin/mesh.wlan.remove
udev/rules.d/50-mesh.wlan.remove.rules

# When tun0 network interface is Added  and trigger /usr/local/sbin/mesh.cjdns.setup
udev/rules.d/50-cjdns.rules
```
(NOTE: hostapad is now merged into 50-mesh.wlan.rules)

The following scripts that are triggered by the above but still hold no configuration at all
```
/usr/local/sbin/mesh.wlan.setup
/usr/local/sbin/mesh.wlan.remove
/usr/local/sbin/mesh.cjdns.setup
```
These scripts read /etc/mesh.conf where all configuration is held
example of mesh.conf

```
[12:42:0d:8a:79:5f]
type=ap
ssid=tomesh
chan=11
wpa=wpa-eap
wpapass=password

[70:f1:1c:03:a9:48]
type=mesh
ssid=tomesh
freq=2412
protocol=80211s
```
Currently this configuration can be changed using mesh.config

benhylau: I guess in this case all those files are part of the OS and mesh-orange will just have:
`(sd_fat_fs_root)/conf.d/05_mesh_conf.tar` unpack into `/etc/mesh.conf`?

darkdrgn2k: In short, yes
in long -> i'll move this to issues #6 configuration management

darkdrgn2k:
In favour of #93

PR #137, opened 2018-01-05
Pin RPI and OPI onboard wireless adapters to wlan0 

darkdrgn2k:
hostapd is hardcoded to wlan0 so this forces onboard driver to be wlan0

### DIFF
--- a/scripts/hostapd/install
+++ b/scripts/hostapd/install
@@ -9,7 +9,7 @@ H_DNS_SERVER_1="fc6e:691e:dfaa:b992:a10a:7b49:5a1a:5e09"
 H_DNS_SERVER_2="fc16:b44c:2bf9:467:8098:51c6:5849:7b4f"
 
 # Create radvd.conf before installing radvd, installation fails without it
-if ! [ -f /etc/radvd.conf ]; then 
+if ! [ -f /etc/radvd.conf ]; then
     sudo cp "$BASE_DIR/radvd.conf" /etc/radvd.conf
 fi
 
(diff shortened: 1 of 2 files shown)


scripts/hostapd/install
```diff
@@ -102,6 +102,17 @@ if [ -f /etc/dhcpcd.conf ]; then
...
 fi
 
+
+# Pin internal wireless adapter to wlan0 on boot
+interfaces=$(ls /sys/class/net/ | grep wlan)
+for interface in $interfaces; do
+    interfaceDriver="$(basename $(readlink /sys/class/net/$interface/device/driver))";
+    if [ "$interfaceDriver" == 'xradio_wlan' ] || [ "$interfaceDriver" == 'brcmfmac' ]; then
+        mac=$(cat /sys/class/net/$interface/address | tr -d '\n')
+	echo "SUBSYSTEM==\"net\", ACTION==\"add\", ATTRS{address}==\"$mac\", KERNEL==\"wlan*\", NAME=\"wlan0\"" |  sudo tee  /etc/udev/rules.d/40-onboard-wlan-pin.rules
```
benhylau: Tab and misalignment

benhylau: What's with the extra spaces around `sudo tee` before and after?

benhylau: The `sudo tee` is wrong this doesn't do anything. You need the `--append` and pipe output to /dev/null please see other places.

benhylau: This is a temporary solution right? We may put USB adapters with these drivers and there would be two `wlan0`?

PR #154, opened 2018-06-25
Added yrd

makeworld-the-better-one:
This also adds a newer IPFS version, which is already done in develop. This was forked from master originally, if you want I can refork from develop.


scripts/install2
```diff
@@ -404,7 +415,10 @@ fi
 if [ ! -x "$(command -v ipfs)" ] && [ ! -z "$WITH_IPFS" -a "$WITH_IPFS" == "true" ]; then
     source ipfs/install
 fi
-
+# yrd
+if [ ! -x "$(command -v ipfs)" ] && [ ! -z "$WITH_YRD" -a "$WITH_YRD" == "true" ]; then
```
benhylau: Remove the ipfs check.

scripts/install2
```diff
@@ -440,6 +454,7 @@ if [[ "$SUPPORT_WATCHDOG" == "true" ]] && [ ! -z "$WITH_WATCHDOG" -a "$WITH_WATC
     source watchdog/install
 fi
 
+
```
benhylau: Why new newline?

scripts/yrd/install
```diff
@@ -0,0 +1,23 @@
...
+
+# get previous location - should be 
+START_DIR="$(pwd)"
+
+BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+mkdir "$BASEDIR/tmp"
+cd "$BASEDIR/tmp"
+git clone https://github.com/kpcyrd/yrd.git
+cd yrd
+# install dependencies
```
benhylau: For consistency with other files in this repo, comments are formatted like this:

`# Install dependencies`

With `#` then space then capital unless it's a name that starts with a lowercase letter (such as yrd).

scripts/yrd/uninstall
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
```
benhylau: See how ipfs/uninstall does it. Should disable, then stop, then daemon-reload, then remove. Also notice the `|| true` and `rm -f` flag. These prevent errors when things aren't running to being with when you uninstall.

```
# Uninstall IPFS
sudo systemctl disable ipfs.service 2>/dev/null || true
sudo systemctl stop ipfs.service 2>/dev/null || true
sudo systemctl daemon-reload
sudo rm -f /usr/local/bin/ipfs
sudo rm -f /etc/systemd/system/ipfs.service
```

scripts/yrd/install
```diff
@@ -0,0 +1,23 @@
...
+sudo apt-get install python3 python3-pip python3-argh python3-requests
+# install yrd
+sudo -H python3 ./setup.py -v install
+# install and enable the service
+cd "$START_DIR"
+sudo cp yrd.service /etc/systemd/system/
+sudo systemctl enable yrd
```
benhylau: Let's add this before enabling to be safe:

```
sudo systemctl daemon-reload
```

Maybe it's no longer necessary but I have had issues with earlier images.

scripts/yrd/install
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -e
+
+# get previous location - should be 
+START_DIR="$(pwd)"
```
benhylau: I think this line should go after the `BASE_DIR` line.

darkdrgn2k: Don't think `START_DIR` is even needed. The place that it is used `BASE_DIR` should have been used for consistency

scripts/yrd/install
```diff
@@ -0,0 +1,23 @@
...
+cd "$BASEDIR/tmp"
+git clone https://github.com/kpcyrd/yrd.git
+cd yrd
+# install dependencies
+sudo apt-get install python3 python3-pip python3-argh python3-requests
+# install yrd
+sudo -H python3 ./setup.py -v install
+# install and enable the service
+cd "$START_DIR"
+sudo cp yrd.service /etc/systemd/system/
```
darkdrgn2k: Don't need to change work directory.   And relaying on work directory is not the correct action as work directory can be different then where the file is located.

Instead use `BASE_DIR`
`sudo cp "$BASE_DIR/yrd.service" "/etc/systemd/system/yrd.service"`

makeworld-the-better-one:
@benhylau @darkdrgn2k I made changes and pushed them to this branch. Can we merge now, or is there another step?

scripts/yrd/install
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -e
+
+BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+mkdir "$BASEDIR/tmp"
+cd "$BASEDIR/tmp"
+git clone https://github.com/kpcyrd/yrd.git
+cd yrd
+# Install dependencies
```
benhylau: Usually we put a blank line above each comment

scripts/yrd/uninstall
```diff
@@ -0,0 +1,11 @@
...
+
+set -e
+
+# Uninstall yrd
+sudo systemctl disable yrd 2>/dev/null || true
+sudo systemctl stop yrd 2>/dev/null || true
+sudo systemctl daemon-reload
+sudo pip3 uninstall yrd
+sudo apt-get remove python3-argh python3-pip python3-requests python3
```
benhylau: This isn't used by other stuff right?

PR #157, opened 2018-06-30
ssb web

darkdrgn2k:


scripts/ipfs/ipfs-http-gateway.conf
```diff
@@ -0,0 +1,15 @@
+    location /ipfs {
```
darkdrgn2k: Previous config was wrapped in a server tag.
Since i need to co-exist/share the same server with multiple module, i created a server tag that includes theses files into it, making one sever tag with multiple modules.

Also changed it from proxying / to individual paths.

scripts/ssb-web/install
```diff
@@ -0,0 +1,57 @@
...
+set -e
+
+BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Install nginx and php
+source "$BASE_DIR/../shared/nginx/install"
+sudo apt-get install -y php-fpm socat
+
+# Install required parts of ssb from npm
+sudo npm install -g --unsafe-perm \
```
ansuz: an alternative here is to install npm with [nvm](https://github.com/creationix/nvm), which allows you to specify which version of npm you'd like to use, and works around needing `sudo` for `-g` packages (they will be run from your user's home directory instead.

If you need npm for multiple users, it may not be suitable out-of-the-box.

darkdrgn2k: npm is WIERD. 

I install it globally but a package but it still complains about the package is missing.
i have to link the global repo into the "users" local repo as a work around

scripts/ssb-web/uninstall
```diff
@@ -0,0 +1,13 @@
...
+sudo rm -rf /var/www/sbot
+sudo rm -rf /usr/lib/node_modules
+sudo rm -rf /etc/nginx/site-path-enabled/ssb-web.conf
+
+sudo npm uninstall -g --unsafe-perm \
+                 ssb-keys \
+                 ssb-client \
+                 ssb-feed \
+                 pull-stream
+sudo apt-get remove -y php-fpm socat
```
makeworld-the-better-one: This script should change the perms back, and remove /var/www/.ssb. /usr/local/bin/sbot should be removed as well, and so should $BASEDIR/tmp/ssb-web-pi, maybe even at the end of the of the install script.

PR #158, opened 2018-07-07
Raspberry ipfs pi stream

darkdrgn2k:
Stream a raspberry pi camera over IPFS


scripts/ipfs-pi-stream/uninstall
```diff
@@ -0,0 +1,14 @@
...
+BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+sudo systemctl stop process-stream
+sudo systemctl disable process-stream
+sudo rm -f /usr/bin/process-stream.sh
+sudo rm -f /etc/systemd/system/process-stream.service
+sudo systemctl daemon-reload
+
+# remove ffmpeg and supporting tools
+sudo apt-get -y remove ffmpeg lsof inotify-tools
```
makeworld-the-better-one: Is there any point it putting back the nginx directory permissions as they were before? I'm thinking no, but I just wanted to bring it up.

scripts/ipfs-pi-stream/install
```diff
@@ -0,0 +1,23 @@
...
+sudo usermod -a -G $USER www-data
+sudo chmod g+rw /var/www/html
+
+# TODO: why is this needed?
+sudo chmod a+rw /var/www/html
+
+sudo cp -f $BASE_DIR/process-stream.sh /usr/bin/process-stream.sh
+sudo cp -f $BASE_DIR/process-stream.service /etc/systemd/system/process-stream.service
+sudo systemctl daemon-reload
+sudo systemctl enable process-stream
```
makeworld-the-better-one: I'm not sure if enabling this is a good idea, since it means a live stream is being constantly recorded since boot, which isn't really made obvious in the install script or the readme. The user can always disable it of course, but I would at least mention it somewhere.

scripts/ipfs-pi-stream/process-stream.sh
```diff
@@ -0,0 +1,112 @@
...
+      echo "#EXT-X-MEDIA-SEQUENCE:${sequence}" >> current.m3u8
+      tail -n ${M3U8_SIZE} ~/process-stream.log | awk '{print $6"#EXTINF:"$5",\n'${IPFS_GATEWAY}'/ipfs/"$2}' | sed 's/#EXT-X-DISCONTINUITY#/#EXT-X-DISCONTINUITY\n#/g' >> current.m3u8
+
+      # Add m3u8 file to IPFS and IPNS publish
+      m3u8hash=$(ipfs add current.m3u8 | awk '{print $2}')
+      ipfs name publish --timeout=5s $m3u8hash &
+
+      # Copy files to web server
+      cp current.m3u8 /var/www/html/live.m3u8
+      cp ~/process-stream.log /var/www/html/live.log
```
makeworld-the-better-one: The installation process for IPFS removes the default nginx config, and just has the IPFS reverse proxy one. This means that these files being copied over won't be served, and even if the default config wasn't removed, you'd still have the reverse proxy competing for port 80 with the default config.

IPFS should use a different port, and the default config should remain if you want the files to be able to be served.

scripts/ipfs-pi-stream/process-stream.sh
```diff
@@ -0,0 +1,112 @@
...
+    attempts=5
+    until [[ ${ret} -eq 0 || ${attempts} -eq 0 ]]; do
+      # Wait and retry
+      sleep 1
+      ret=`ipfs add ${nextfile} 2>/dev/null > ~/tmp.txt; echo $?`
+      attempts=$((attempts-1))
+    done
+    if [[ ${ret} -eq 0 ]]; then
+      # Update the log with the future name (hash already there)
+      echo $(cat ~/tmp.txt) ${time}.ts ${timecode}${reset_stream_marker} >> ~/process-stream.log
```
makeworld-the-better-one: A small thing, you could just do `ipfs add -Q ${nextfile}`, which will only give the hash. That might also be useful for the log, as `ipfs add ${nextfile} 2>/dev/null > ~/tmp.txt` means tmp.txt will read `added $hash $filename`, but maybe you already have the file name: `echo $(cat ~/tmp.txt) ****${time}.ts**** ${timecode}${reset_stream_marker} >> ~/process-stream.log`

darkdrgn2k: log is only a log of the hashes in order. Its used on line 99 to assemble the m3u8 file

If you add errors into this file you will end up with a mess.  However errors will be put in /var/log/syslog since it is daemon

scripts/ipfs-pi-stream/process-stream.sh
darkdrgn2k: Issue is in a new share model of nginx in another pull request.

scripts/ipfs-pi-stream/install
darkdrgn2k: You also get a nice bright light in front of the camera telling you its recording.  I wanted a 0 fuss setup of this.

However adding it to the docs does make sense i think

scripts/ipfs-pi-stream/process-stream.sh
makeworld-the-better-one: I wasn't talking about errors. I'm saying how logging the output `ipfs add ${somefile}` might be more redundant then you need, because it will say `added $hash $filename`, but maybe all you wanted to log was `$hash`.

scripts/ipfs-pi-stream/install
makeworld-the-better-one: I didn't know about the light, and I understand that this isn't the final, secure, deployment product. Adding it to the docs sounds good to me!

scripts/ipfs-pi-stream/process-stream.sh
makeworld-the-better-one: The ssb-web one right? I just saw that after I published the review.

scripts/ipfs-pi-stream/process-stream.sh
darkdrgn2k: `ipfs add -Q ${nextfile}` is a great solution, however im not going to change it for now since this mirrors what is done in ipfs-live-stream.

scripts/ipfs-pi-stream/process-stream.sh
```diff
@@ -0,0 +1,112 @@
...
+          sequence=`expr ${totalLines} - ${M3U8_SIZE}`
+      fi
+      echo "#EXTM3U" > current.m3u8
+      echo "#EXT-X-VERSION:3" >> current.m3u8
+      echo "#EXT-X-TARGETDURATION:${HLS_TIME}" >> current.m3u8
+      echo "#EXT-X-MEDIA-SEQUENCE:${sequence}" >> current.m3u8
+      tail -n ${M3U8_SIZE} ~/process-stream.log | awk '{print $6"#EXTINF:"$5",\n'${IPFS_GATEWAY}'/ipfs/"$2}' | sed 's/#EXT-X-DISCONTINUITY#/#EXT-X-DISCONTINUITY\n#/g' >> current.m3u8
+
+      # Add m3u8 file to IPFS and IPNS publish
+      m3u8hash=$(ipfs add current.m3u8 | awk '{print $2}')
```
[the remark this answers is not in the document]
darkdrgn2k: That would be a pretty long log as every chunkn would have an ipns publis with it.

scripts/ipfs-pi-stream/process-stream.sh
darkdrgn2k: :+1:

makeworld-the-better-one: Even though ssb-web solves this, this will still have to be changed to work with the new model so that these files can be distributed.

PR #163, opened 2018-07-22
Peer ipfs bootstrap

makeworld-the-better-one:
This branch adds a module that checks to see if ipfs has successfully boostrapped and has peers, and if it doesn't, it looks at the node's cjdns peers to see if any of them are running ipfs, and can be bootstrapped off of. If they can, it adds those ip addresses to the bootstrap list, and restarts ipfs. As of right now, it only looks at the node's 1-hop cjdns peers, because the command cjd gave me to look at the peers of peers (ie looking beyond 1-hop) doesn't work. 

This branch also introduces a nodeinfo.json system for nodes to share info about what modules they have enabled, among other things. For now only IPFS is listed, but in theory every module will be registered in this file so other nodes can know who's running what. This file is served from /var/www/files, which I created so that any file can be put in there as a quick web server, when using a distributed system like IPFS doesn't make sense. 

This branch makes use of jq to process json. This could be used further clean up a lot of the awk and sed usage that happens for json, for example getting stuff cjdroute.conf.

Also, this is untested. I plan to test it soon, but I wanted to make the PR first.


scripts/install2
```diff
@@ -507,6 +508,26 @@ fi
 sudo cp status /usr/local/bin/status
 echo -e "Run \e[1;32mstatus\e[0m anytime to print the status of your node"
 
+# Setup nodeinfo.json, adding non-static info
+echo '"installed":'"$(date)"', "uri":"https://github.com/tomeshnet/prototype-cjdns-pi/tree/develop"},' >> shared/nodeinfo/nodeinfo.json
```
makeworld-the-better-one: When we eventually have a versioning system, that could go here along with the date. Also note that that uri will need to change when this gets merged to master.

scripts/peer-ipfs-bootstrap/uninstall
```diff
@@ -0,0 +1,14 @@
...
+
+# Remove script and service
+sudo systemctl disable peer-ipfs-bootstrap.service 2>/dev/null || true
+sudo systemctl stop peer-ipfs-bootstrap.service 2>/dev/null || true
+sudo systemctl daemon-reload
+sudo rm -r /usr/local/bin/check-and-add-peers.sh
+sudo rm -f /etc/systemd/system/peer-ipfs-bootstrap.service
+
+# Uninstall dependencies
+sudo apt-get remove jq
```
makeworld-the-better-one: Maybe I should leave this installed in case it's used in the future?

scripts/shared/nodeinfo/nodeinfo.json
```diff
@@ -0,0 +1,11 @@
...
+    "location": {
+        "continent": "NA",
+        "country": "Canada",
+        "region": "Ontario",
+        "municipality": "Toronto",
+        "uri": "https://tomesh.net"
+    },
+    "software": {
+        "repo": "prototype-cjdns-pi",
+        "branch": "develop",
```
makeworld-the-better-one: This will also need to be changed when develop gets merged into master.

scripts/install2
darkdrgn2k: Why are we modifying local files  on install?

scripts/ipfs/uninstall
```diff
@@ -25,3 +25,6 @@ fi
 sudo rm -f /etc/nginx/sites-enabled/ipfs-http-gateway
 sudo rm -f /etc/nginx/sites-available/ipfs-http-gateway
 sudo systemctl restart nginx.service
+
+# Remove a dependant module
+source ../peer-ipfs-bootstrap/uninstall
```
darkdrgn2k: We don't normally remove "Depended" modules since i don't think we support removing a single module

scripts/shared/nginx/main.conf
```diff
@@ -2,4 +2,9 @@ server {
   listen 80;
   listen [::]:80;
   include /etc/nginx/site-path-enabled/*.conf;
+  
+  root /var/www/files;
```
darkdrgn2k: I'd leave it as `root /var/www/html/` as this is the default

scripts/shared/nginx/main.conf
```diff
@@ -2,4 +2,9 @@ server {
...
+  location / {
```
darkdrgn2k: I would not auto index
and if you don't auto index you don't need a /location

scripts/shared/nodeinfo/nodeinfo.json
darkdrgn2k: Changing this when being merged will be hard. Instead consider:

repo ` git remote get-url origin | awk  -F / '{print $5}'|  cut -d '.' -f1`
branch: `git branch | grep \* | cut -d ' ' -f2` or  `git rev-parse --abbrev-ref HEAD`
maybe also add current commit `git rev-parse HEAD`

makeworld-the-better-one: Will do.

scripts/install2
```diff
@@ -449,6 +449,7 @@ fi
 # IPFS
 if [ ! -x "$(command -v ipfs)" ] && [ ! -z "$WITH_IPFS" -a "$WITH_IPFS" == "true" ]; then
     source ipfs/install
+    source peer-ipfs-bootstrap/install
```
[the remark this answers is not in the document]
makeworld-the-better-one: But shouldn't it just be installed along with IPFS? Or you'd rather it have a dialog in the install script asking whether it should be installed?

scripts/install2
makeworld-the-better-one: To add dynamic things to the json file, like node ids and things. I realize now that just putting it in the nodeinfo install and running that is better.

scripts/shared/nginx/main.conf
makeworld-the-better-one: Doesn't it apply to more than html though? It doesn't really matter in the end, but I feel like html would be a misnomer.

scripts/shared/nginx/main.conf
makeworld-the-better-one: Auto-indexing allows folders to be served by nginx without error. It will instead provide a simple file browser, showing everything in that directory. This is so that you can add a folder to the /var/www/files intuitively, without having to provide index.html file for each one so that accessing the folder directly doesn't give an error.

makeworld-the-better-one:
All changes have been added, except the one about this being its own module, instead of just being installed alongside ipfs, and about removing and tracking peers as mentioned on matrix. I will work more on the latter, and when @darkdrgn2k has time we can talk about the former.

scripts/peer-ipfs-bootstrap/check-and-add-peers.sh
```diff
@@ -0,0 +1,85 @@
...
+                    else
+
+                        # IPFS has been removed from nodeinfo manually. It is assumed the peer will never be a candidate for bootstrapping again
+                        # The peer is removed
+                        ipfs bootstrap rm $peer
+                        sed -i "/$peer/d" /var/lib/peer-ipfs-bootstrap/peers.data
+                    fi
+                else
+
+                    # It's not online. The peer is removed, since this script does not track long term uptime stats of nodes.
```
makeworld-the-better-one: Should this happen? I figured that if a node is offline, it would be wasteful to keep it, since it may be like that permanently. The only situation where this becomes a problem is if all the nodes you use to connect to IPFS are offline at runtime, and none of them are your 1-hop peers.

scripts/install2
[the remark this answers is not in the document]
benhylau: I think @darkdrgn2k's point is to have the contents of `peer-ipfs-bootstrap/install` in the same file as `ipfs/install` which I agree, as other `WITH_*` components have a 1:1 mapping to the directory.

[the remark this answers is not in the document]
makeworld-the-better-one: Should all the other files be in there as well then, service and script, etc.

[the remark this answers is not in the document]
darkdrgn2k: Yes keep it compartmentalized.

scripts/shared/nodeinfo/nodeinfo.json
```diff
@@ -0,0 +1,9 @@
+{
+    "location": {
+        "continent": "NA",
+        "country": "Canada",
+        "region": "Ontario",
+        "municipality": "Toronto",
+        "uri": "https://tomesh.net"
+    },
+    "software": {
```
benhylau: @makeworld-the-better-one It's a really odd way to do this with a partial json file that isn't actually valid json. It's better to have a properly terminated valid json here, use macros like `__REPLACE_ME__` if needed, then use `jq` to add additional elements.

makeworld-the-better-one: Ok, will do.

scripts/shared/nginx/main.conf
makeworld-the-better-one: Nvm, I've removed it now.

scripts/install2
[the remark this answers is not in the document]
makeworld-the-better-one: Done.

scripts/shared/nodeinfo/nodeinfo.json
makeworld-the-better-one: Done as well, I've used sed to replace stuff though, if that's ok.

PR #168, opened 2018-07-28
SSB fixes for Broadcast

darkdrgn2k:
Reverted ID back to derived instead of passed


benhylau:
No this would not work, we need to grab the ID _in_ the loop and validate whether it's empty.

scripts/ssb/ssb-broadcast-service.sh
```diff
@@ -1,12 +1,14 @@
...
-id="$1"
 
 while true; do
-    for int in $(ls -1Atu /sys/class/net ); do
-        ip=$(ip addr show $int | grep -v inet6 | grep -v '127.0.0.1' |grep inet | head -n 1 | awk '{print $2}' | awk -F "/" '{print $1}')
-        if ! [ -z "$ip" ]; then
-            echo -n "net:$ip:8008~shs:$id" |  sudo socat -T 1 - UDP4-DATAGRAM:255.255.255.255:8008,broadcast,so-bindtodevice=$int &
-        fi
+    id=$(sbot whoami | grep id | awk -F "\"" '{print $4}' | sed 's/.ed25519//' | sed 's/@//')
+    if ! [ -z "$id" ]; then
```
benhylau: There is no closing of the `if`

PR #169, opened 2018-07-28
Fixes for develop

darkdrgn2k:
bunch of errors fixed in develop that break build


makeworld-the-better-one:
Looks good! I won't edit in place next time.

PR #171, opened 2018-07-28
Develop cleanup PR

darkdrgn2k:
Syntax, white space, formatting, and other small issue resolves

Target Full clean install target on

 - [x] Raspberry Pi 3
 - [ ] Orange Pi Zero


scripts/ipfs-pi-stream/process-stream.service
```diff
@@ -8,7 +8,7 @@ Type=simple
 User=pi
 Group=pi
 ExecStart=/usr/bin/process-stream.sh
-ExecStop=/usr/bin/pkill -f process-stream.sh; /usr/bin/pkill -f ffmpeg
+ExecStop=ExecStop=/bin/kill -s QUIT $MAINPID; /usr/bin/pkill -f ffmpeg
```
benhylau: Copy error

scripts/ipfs/install
```diff
@@ -31,3 +31,22 @@ sudo systemctl restart nginx.service
 
 # add cjdns bootstrap address
 ipfs bootstrap add /ip6/fc6e:691e:dfaa:b992:a10a:7b49:5a1a:5e09/tcp/4001/ipfs/QmU6NeD2Uu34WKest1NZGvGmScLhN1zVo66K35GeE6Jft2
```
benhylau: Which node is this? If it's a central node on the Internet, can we have this as a var defined at top of file and add comment describing?

scripts/shared/nodeinfo/nodeinfo.json
```diff
@@ -7,15 +7,15 @@
...
-        "installed": "INSTALLED",
+        "repo": "__REPO__",
+        "branch": "__BRANCH__",
+        "commit": "__COMMIT__",
+        "installed": "__INSTALLED__",
         "uri": "https://github.com/tomeshnet/prototype-cjdns-pi/"
     },
-    "key": "KEY",
-    "org": "ORG",
+    "key": "__KEY__",
```
benhylau: Making cjdns first class makes it difficult to add another router / identity (e.g. yggdrasil, libp2p) in the future. Perhaps:

```
"routers": {
    "cjdns": {
        "key": "__KEY__"
    }
}
```

We don't have to address this now, since this is prototype and we can make breaking changes. Perhaps we need to version this JSON though.

scripts/shared/nodeinfo/nodeinfo.json
```diff
@@ -7,15 +7,15 @@
...
+    "org": "__ORG__",
     "services": {
-        SERVICES
+        __SERVICES__
```
benhylau: It's better to have stable keys for validation and navigation purposes. Like:

```
"services": [
    {
        "name": "ipfs",
        "version": "__IPFS_VERSION__",
        "params": [__IPFS_PARAMS__]
    }
]
```

This is a very rushed schema design. My take is it will completely change in the future, so stick a version on the schema.

scripts/ssb-web/ssb-web-broadcast-service.sh
```diff
@@ -1,13 +1,13 @@
...
-      ip=$(ip addr show $int | grep -v inet6 | grep -v '127.0.0.1' |grep inet | head -n 1 | awk '{print $2}' | awk -F "/" '{print $1}')
-      if ! [ -z "$ip" ]; then
-        echo -n "net:$ip:8008~shs:$id" | sudo socat -T 1 - UDP4-DATAGRAM:255.255.255.255:8008,broadcast,so-bindtodevice=$int &        
-      fi
-    done
-  done <<< $(sudo cat /var/www/backend/keys/* | grep id | grep -v "#" | awk '{print $2}' | tr -d '"' | sed 's/.ed25519//' | sed 's/@//')
-  sleep 15
+    while read -r id; do
+        for int in $(ls -1Atu /sys/class/net ); do
+        ip=$(ip addr show $int | grep -v inet6 | grep -v '127.0.0.1' |grep inet | head -n 1 | awk '{print $2}' | awk -F "/" '{print $1}')
```
benhylau: Indent looks wrong.

scripts/ssb-web/ssb-web.conf
```diff
@@ -1,12 +1,12 @@
-  location /sbot {
-    alias /var/www/sbot;
+location /sbot {
+     alias /var/www/sbot;
```
benhylau: Indent looks wrong.

scripts/ipfs/install
darkdrgn2k: This is heavy metals node that allows us to discover other hyperbole nodes on hyperbole only connections

scripts/shared/nodeinfo/nodeinfo.json
darkdrgn2k: My bigger question is along side node exporter

How much do we want to share

scripts/ipfs/install
makeworld-the-better-one: @benhylau I think it's one of @darkdrgn2k 's servers, the one which also serves as hyperboria DNS server.

darkdrgn2k: Oh i guess so i though heavy metal ran that one.  Guess not. Added comment

scripts/shared/nodeinfo/nodeinfo.json
darkdrgn2k: Created issue #173

scripts/shared/nodeinfo/nodeinfo.json
darkdrgn2k: Also rolled into issue #173

scripts/shared/nodeinfo/nodeinfo.json
makeworld-the-better-one: @benhylau Good point. I was trying to stick to the standard outlined [here](https://docs.meshwith.me/cjdns/nodeinfo-json.html) to some degree, but there's not a good reason why we have to.

scripts/shared/nodeinfo/nodeinfo.json
makeworld-the-better-one: @benhylau That is what the original nodeinfo spec said to do. Would you rather this be changed before the release? And where should the version be recorded?

scripts/install2
```diff
@@ -377,15 +352,17 @@ if [[ ! -e "/usr/bin/cjdroute" ]]; then
...
-        sudo rm -rf /opt/cjdns
+    if ! [ -x "/opt/cjdns/cjdroute" ]; then
+        echo -e "\e[1;31mInstalling cjdns Debian package\e[0m"
+        wget https://github.com/darkdrgn2k/ToMeshStuff/raw/master/deb/$CJDNS_PACKAGE -O /tmp/cjdns.deb
+        sudo dpkg -i /tmp/cjdns.deb
+        #Link up folder so files are in the same place as compiled
+        if [[ -e "/opt/cjdns" ]]; then
+            sudo rm -rf /opt/cjdns
+        fi
+        sudo ln -s /usr/share/cjdns /opt/cjdns
```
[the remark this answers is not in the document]
darkdrgn2k: yes 
but if it has not, the script will crash so its to add resilience.

scripts/ipfs/peer-ipfs-bootstrap/check-and-add-peers.sh
```diff
@@ -1,86 +0,0 @@
-#!/bin/bash
```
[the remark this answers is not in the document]
darkdrgn2k: Conflict and i guess it didnt get put back,

scripts/install2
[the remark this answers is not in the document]
makeworld-the-better-one: No I mean, how can a symbolic link be made to a folder that doesn't exist?

scripts/shared/nodeinfo/nodeinfo.json
benhylau: The only thing we need to change in this release is add:

`"version": 0,`

to the top level, as a sibling to `"services"`. This will allow us to detect breaking changes and since this is prototype, we will break backwards compatibility as needed :)

My point is, schema design is something that will take a long time to discuss, and even then it will evolve. Adding a version is absolutely a smart thing to do now.

scripts/install2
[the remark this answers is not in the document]
benhylau: >but if it has not, the script will crash so its to add resilience.

Huh? I don't understand this logic. If we want the file gone, we want to be sure it's gone, and we should not expect that it _may_ be there. And if the file is gone, why do we `ln` it?

scripts/shared/nodeinfo/nodeinfo.json
darkdrgn2k: Added in https://github.com/tomeshnet/prototype-cjdns-pi/commit/3e8e30fe2dadbb72362d43d59a45bdb3c58b44d2

scripts/ipfs/install
```diff
@@ -31,3 +31,22 @@ sudo systemctl restart nginx.service
...
+
+# Service to bootstrap neighboring cjdns nodes if no other ipfs is found
```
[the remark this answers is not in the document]
darkdrgn2k: As i said before. If its part of ipfs keep it part of ipfs
if its a separate module keep it a separate module

i though we agreed on keeping it as part of one module?

scripts/install2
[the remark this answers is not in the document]
benhylau: I understand why you remove a previous installation, but I don't see how this explains the line that follows:

>sudo ln -s /usr/share/cjdns /opt/cjdns

I am as confused as @makeworld-the-better-one about this.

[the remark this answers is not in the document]
darkdrgn2k: Deb package installs binaries into /usr/share/cjdns
Prototype scripts look files in /opt/cjdns

therefore

`sudo rm -rf /opt/cjdns` removes compiled/linked copy of cjdns
`sudo ln -s /usr/share/cjdns /opt/cjdns` creates the links to the deb packages location

The change here is simply removing the existence of existing installation

darkdrgn2k:
> I thought we were doing it as a submodule, where it is a separate folder, but gets installed automatically by the parent module.

Trying to keep modules contained similarly to what it would look like if they where Debian packages. The install trigger only happens once during a deb install.

scripts/shared/nodeinfo/nodeinfo-update.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+cp /opt/tomesh/nodeinfo.json /tmp
+
+# Replace placeholders with dynamic info
+sed -i -e "s/__KEY__/$(grep -m 1 '"ipv6"' /etc/cjdroute.conf | awk '{ print $2 }' | sed 's/[",]//g')/g" /tmp/nodeinfo.json
```
makeworld-the-better-one: Why isn't __KEY__ in the install file?

darkdrgn2k: Because CJDNS key CAN change (unlikely but can) and we can easy extrapolate it on boot so why not.

Git data that is part of the install is a little harder since the git repo can be deleted after the install, and we don't know WHERE it was downloaded to.

scripts/ipfs/nodeinfo-ipfs
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
```
benhylau: `.sh`?

darkdrgn2k: bug in run-parts that does not executable .sh files.

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -0,0 +1,36 @@
...
+while read -r cjdns_peer; do
+    cjdns_addr=$(sudo /opt/cjdns/publictoip6 $cjdns_peer)
+
+    # See if they have IPFS enabled
+    res=$(curl http://[${cjdns_addr}]/nodeinfo.json)
+    if [ ! -x "$res" ]; then
+        id=$(echo ${res} | jq -r -M '.services.ipfs.ID')
+        # Value is found
+        if [[ ! ${id} == "null" ]] && [[ ! "${id}" == "" ]]; then
+            # Add them as a bootstrap peer
```
makeworld-the-better-one: I'm being annoying, but *technically* they are being connected to, not added as a bootstrap peer.

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -0,0 +1,36 @@
...
+    fi
+
+    # Add all that node's peers to the bottom of the list to check further hop peers
+    # XXX: The below command hasn't been working -- so for now only 1-hop peers are checked
+    #peers+=$(cjdnstool query getpeers $peer | sed -e '1d;$d' |awk -F. '{ print $6".k" }')
+
+done <<< `sudo nodejs /opt/cjdns/tools/peerStats 2>/dev/null | awk '{ if ($2 == "ESTABLISHED") print $1 }' | awk -F. '{ print $6".k" }' | xargs`
+
+# update peers data since ipfs just started
+/usr/local/bin/nodeinfo-update.sh
```
makeworld-the-better-one: What peer data is being shared by IPFS? I can't see any.

darkdrgn2k:
Building on orange pi and raspberry pi tonight, if all goes well I will merge in the morning to develop and then start a new pr to iron out anything else we may find in prep for release.

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -2,23 +2,23 @@
...
+    res=$(curl http://[${cjdns_addr}]/nodeinfo.json)
+    if [ ! -x "$res" ]; then
         id=$(echo ${res} | jq -r -M '.services.ipfs.ID')
         # Value is found
         if [[ ! ${id} == "null" ]] && [[ ! "${id}" == "" ]]; then
             # Add them as a bootstrap peer
-            ipfs swarm connect "/ip6/${peer}/tcp/4001/ipfs/${id}"
-            echo "Connecting to ${peer}."
+            ipfs swarm connect "/ip6/${cjdns_addr}/tcp/4001/ipfs/${id}"
+            echo "Connecting to ${cjdns_addr}."
```
benhylau: Remove the `.` at the end? We don't do periods in all the other messages

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -28,5 +28,5 @@ while read -r peer; do
 
 done <<< `sudo nodejs /opt/cjdns/tools/peerStats 2>/dev/null | awk '{ if ($2 == "ESTABLISHED") print $1 }' | awk -F. '{ print $6".k" }' | xargs`
 
-# update peers data since ipfs just started 
+# update peers data since ipfs just started
```
benhylau: Capitalize start of comment

PR #172, opened 2018-07-29
[WIP] nodeinfo.d

darkdrgn2k:
New modular method of generating and update nodeinfo,json

WIP To do
- [x] Add spawning nodeinfo update on boot 
- [x] Confirm ipfs module runs as proper user
- [x] Confirm working correctly on clean install
- [x] Correct syntax and spacing


scripts/shared/nodeinfo/nodeinfo-update.sh
```diff
@@ -0,0 +1,17 @@
...
+sed -i -e "s/__REPO__/$(git remote get-url origin | awk -F / '{print $5}'| cut -d '.' -f1)/g" /tmp/nodeinfo.json
+sed -i -e "s/__BRANCH__/$(git rev-parse --abbrev-ref HEAD)/g" /tmp/nodeinfo.json
+sed -i -e "s/__COMMIT__/$(git rev-parse HEAD)/g" /tmp/nodeinfo.json
+sed -i -e "s/__INSTALLED__/$(date)/g" /tmp/nodeinfo.json
+
+sed -i -e "s/__KEY__/$(grep -m 1 '"ipv6"' /etc/cjdroute.conf | awk '{ print $2 }' | sed 's/[",]//g')/g" /tmp/nodeinfo.json
+sed -i -e "s/__ORG__/$MESH_NAME/g" /tmp/nodeinfo.json
+
+services=$(run-parts /opt/tomesh/nodeinfo.d/ | sed ':a $!{N; ba}; s/\n/\\n/g')
+sed -i -e "s/__SERVICES__/$services/g" /tmp/nodeinfo.json
```
benhylau: Can we `jq` instead of `sed` and not use macros here?

darkdrgn2k: For this pr followed make world lead 
Question is  favour jq over sed everywhere or less dependencies

Also jq does not work with cjdouter.conf because of comments

makeworld-the-better-one: @darkdrgn2k @benhylau I feel like since jq is already a dependency it should be used where it's possible. If it doesn't work with cjdroute.conf that sucks, but when it can be used, it should, because it's much easier to write and read.

darkdrgn2k: Just a note - JQ is actually NOT a dependency except for this particular instance.

It was removed in other instance because it didn't work and would not work.

scripts/ipfs/peer-ipfs-bootstrap/check-and-add-peers.sh
```diff
@@ -0,0 +1,80 @@
...
+                fi
+
+                # Remove them
+                peers=("${peers[@]/$peer}")
+
+                # Add all that node's peers to the bottom of the list to check further hop peers
+                # XXX: The below command hasn't been working -- so for now only 1-hop peers are checked
+                #peers+=$(cjdnstool query getpeers $peer | sed -e '1d;$d' |awk -F. '{ print $6".k" }')
+
+        done <<< `sudo nodejs /opt/cjdns/tools/peerStats 2>/dev/null | awk '{ if ($2 == "ESTABLISHED") print $1 }' | awk -F. '{ print $6".k" }' | xargs`
```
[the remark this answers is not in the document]
darkdrgn2k: This code has changed

Your code
````
read -a peers <<< `sudo nodejs /opt/cjdns/tools/peerStats 2>/dev/null | awk '{ if ($2 == "ESTABLISHED") print $1 }' | awk -F. '{ print $6".k" }' | xargs`
```
Read all contents into a variable


This is a "foreach" loop

Start of loop (for each as $peers)
```
while read -r peers; do
```

End of loop ( Here is the actual iteration content)
```
```

scripts/ipfs/peer-ipfs-bootstrap/check-and-add-peers.sh
```diff
@@ -0,0 +1,80 @@
...
+
+        echo "Restarting ipfs.service..."
```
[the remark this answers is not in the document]
darkdrgn2k: Don't really need to restart IPFS after add bootstrap
So no need to keep track of new_peers

scripts/ipfs/peer-ipfs-bootstrap/check-and-add-peers.sh
```diff
@@ -0,0 +1,80 @@
...
+
+                    # It's not online. The peer is removed, since this script does not track long term uptime stats of nodes.
+                    ipfs bootstrap rm $peer
+                    sed -i "/$peer/d" /var/lib/peer-ipfs-bootstrap/peers.data
+                fi
+
+            done
+        fi
+
+        while read -r peers; do
```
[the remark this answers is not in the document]
darkdrgn2k: No need to keep track of # of new peers
Also see explanation of bash for each loop below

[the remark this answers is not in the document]
makeworld-the-better-one: I understand now, but shouldn't this be `read -r peer` then, and the line `peer=${peers[0]}` can be removed?

scripts/ipfs/peer-ipfs-bootstrap/check-and-add-peers.sh
```diff
@@ -0,0 +1,76 @@
...
+
+                        # Add them as a bootstrap peer
+                        ipfs bootstrap add "/ip6/${peer}/tcp/4001/ipfs/${id}"
+                        echo "/ip6/${peer}/tcp/4001/ipfs/${id}" >> /var/lib/peer-ipfs-bootstrap/peers.data
+                        echo "Added cjdns peer ${peer} as a bootstrap node for IPFS."
+                    fi
+                fi
+
+                # Remove them
+                peers=("${peers[@]/$peer}")
```
makeworld-the-better-one: This line could also be removed then.

Issue #173, opened 2018-07-29
Node Exporter and NodeInfo


darkdrgn2k:
- What do we would like nodeexporter (or nodeexporter replacement) to report.
- What kind of information do we want nodeinfo.json to provide
- syntax of nodeinfo.json -  [refrence](https://github.com/hyperboria/docs/blob/master/cjdns/nodeinfo-json.md)

darkdrgn2k:
C/P From PRs


benhylau:
---
Making cjdns first class makes it difficult to add another router / identity (e.g. yggdrasil, libp2p) in the future. Perhaps:

```
"routers": {
    "cjdns": {
        "key": "__KEY__"
    }
}
```

We don't have to address this now, since this is prototype and we can make breaking changes. Perhaps we need to version this JSON though.


benhylau: 
---
It's better to have stable keys for validation and navigation purposes. Like:

```
"services": [
    {
        "name": "ipfs",
        "version": "__IPFS_VERSION__",
        "params": [__IPFS_PARAMS__]
    }
]
```

This is a very rushed schema design. My take is it will completely change in the future, so stick a version on the schema.

makeworld-the-better-one:
I feel like nodeexporter should handle live stats, and the tracking of those stats over periods of time, etc. Nodeinfo should hold mostly static information that's important to have accessible, like the software version nodes are running, or their IPFS ids.

benhylau:
I did not realize there is a somewhat official spec. My comments are consistent with the official spec. The official one looks reasonable if it adds a `"version"` field, I don't know why we decided to implement this other schema.

makeworld-the-better-one:
@benhylau I agree with you that sticking to the spec is best. I didn't do it at first, but I see how it makes more sense now. And yes a version field would be great as well.

darkdrgn2k:
Q: Is exposing IPFS ID to the world a smart idea? Are there security implications to this?

makeworld-the-better-one:
I believe IPFS nodes exchange ids anyway, verifying that they match public keys and things. I don't think it should be an issue.

makeworld-the-better-one:
Yes I believe that may be possible as well, but it's still necessary. Cause to add a node as bootstrap you need that ID anyway, and you can't connect to it without the ID, so you can't probe any handshake.

benhylau:
I don't think it is an issue to expose IPFS ID. They are publicly shared already. However, I think it's important that we don't end up with stale info on the json.

[...]

PR #178, opened 2018-08-13
Develop - ipfs bootstrap

darkdrgn2k:
Finally working code for ipfs ~~~bootstrap~~~ swarm connect.

- removed the need for a separate service by using `ExecStartPost` of ipfs instead
- changed ipfs id to using curl via local api, solves issue of having to have fs permissions and locations
- renamed `check-and-add-peers.sh` to `ipfs-swarm-cjdns.sh` more descriptive name
- use `ipfs swarm connect` instead of bootstrap, this way we don't have to worry about keeping track of badly added bootstraps
- Added wait for ipfs to actually start

More testing required but this does seem to work!


scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -0,0 +1,34 @@
...
+
+while read -r peer; do
+
+    peer=$(sudo /opt/cjdns/publictoip6 $peer)
+
+    # See if they have IPFS enabled
+    res=$(curl http://[${peer}]/nodeinfo.json)
+    if [ ! -x "$res" ]; then
+            
+        id=$(echo ${res} | jq -r -M '.services.ipfs.ID')
```
benhylau: There really should be a shared module that handles schema version checking and parsing, instead of having each module `jq` parse this. In the current model, it means once the schema updates, every module will need to update its `nodeinfo` parsing.

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -0,0 +1,34 @@
...
+
+# Wait for ipfs to initalize
+attempts=15
+until [[ $( curl http://localhost:5001/api/v0/id -q 2>/dev/null) || ${attempts} -eq 0 ]]; do
+    sleep 1
+    attempts=$((attempts-1))
+done
+
+while read -r peer; do
+
```
benhylau: Can we please remove all these random newlines after a `while` and `if`? @darkdrgn2k you do this a lot, but also not all the time. It makes the code very inconsistent.

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -0,0 +1,34 @@
...
+    peer=$(sudo /opt/cjdns/publictoip6 $peer)
+
+    # See if they have IPFS enabled
+    res=$(curl http://[${peer}]/nodeinfo.json)
+    if [ ! -x "$res" ]; then
```
benhylau: `${res}`

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+
+# Wait for ipfs to initalize
+attempts=15
+until [[ $( curl http://localhost:5001/api/v0/id -q 2>/dev/null) || ${attempts} -eq 0 ]]; do
+    sleep 1
+    attempts=$((attempts-1))
+done
+
+while read -r peer; do
```
benhylau: Rename `peer` to something descriptive like `cjdns_ipv6`?

scripts/ipfs/nodeinfo-ipfs
```diff
@@ -1,11 +1,18 @@
 #!/bin/sh
```
benhylau: `.sh` ext?

scripts/ipfs/ipfs-swarm-cjdns.sh
darkdrgn2k: comment added to #173 
Leaving it for now

scripts/ipfs/nodeinfo-ipfs
darkdrgn2k: It was removed because run-parts does not work with .sh extensions.... *shrug*

simply does not run them if it has a .sh ext.

scripts/ipfs/install
```diff
@@ -20,40 +20,27 @@ rm -rf "$BASE_DIR/tmp"
...
-sudo touch /var/lib/peer-ipfs-bootstrap/peers.data
-sudo chmod 777 /var/lib/peer-ipfs-bootstrap/peers.data
+sudo cp $BASE_DIR/ipfs-swarm-cjdns.sh /usr/local/bin/
+sudo chmod +x /usr/local/bin/ipfs-swarm-cjdns.sh
 
-# Enable service
-sudo cp $BASE_DIR/peer-ipfs-bootstrap/peer-ipfs-bootstrap.service /etc/systemd/system
+# Configure systemd to start ipfs.service on system boot
+sudo cp "$BASE_DIR/ipfs.service" /etc/systemd/system/ipfs.service
+sudo sed -i "s|__USER_HOME__|${HOME}|" /etc/systemd/system/ipfs.service
```
jedahan: Totally a style thing, but I've been doing stuff like:

```bash
sed -i "s|__USER_HOME__|${HOME}|" "$BASE_DIR/ipfs.service.template" | sudo tee /etc/systemd/system/ipfs.service
```

Which can also be written as 

```bash
sed -i "s|__USER_HOME__|${HOME}|" "$BASE_DIR/ipfs.service.template" | sudo dd of=/etc/systemd/system/ipfs.service
```

If you don't care about tee copying to stdout

scripts/shared/nodeinfo/nodeinfo-update.sh
```diff
@@ -8,6 +8,6 @@ sed -i -e "s/__KEY__/$(grep -m 1 '"ipv6"' /etc/cjdroute.conf | awk '{ print $2 }
 services=$(run-parts /opt/tomesh/nodeinfo.d/ | sed ':a $!{N; ba}; s/\n/\\n/g')
 services=${services%,*}
 
-sed -i -e "s/__SERVICES__/$services/g" /tmp/nodeinfo.json
+sed -i -e "s#__SERVICES__#$services#g" /tmp/nodeinfo.json
```
jedahan: why not `|` instead of `#` like above?

scripts/ipfs/install
darkdrgn2k: Looks more complicated? or am I missing something.

darkdrgn2k: I understand what you mean now, and yes i agree with you. This was a poorly coded.

I will make an issue.

PR #179, opened 2018-08-14
Release Candidate v3

darkdrgn2k:
## ToDo
- [x] Full Clean install on Raspberry pi 3(b+)
- [ ] Full Clean install on Raspberry pi 2
- [x] Full Clean install on Orange Pi Zero
- [x] Full Clean install on Nano Pi Neo2  (64 bit)

- [x] Check IPFS cjdns peering


darkdrgn2k:
Release summary

## FAQ
- Note about hardcoded machine ids on some armbian distros

## Modules
- Added yrd Module
- Added ssb module
- Added ssb web Module
- Bump nodejs from 7.x to 8.x 
- Corrected binary package logic for cjdns 
- New shared module framework
- Nodesinfo module V0
- Added ipfs-swarm-cjdns.sh to connect cjdns neighbors to swarm
- Enabled  namesys-pubsub and migrate modules for IPFS
- IPFS bump from v0.4.6 to v0.4.17
- NodeExporter reports mac addresses for mapping

## Hardware/OS
- Added rfkill on boot to unblock wifi on some Raspberry Pi 
- Removed xradio hack for older kernel
- Added sun50iw1p1 to detect some 64 arm boards (armbian kernel update)
- Compile cjdns with -O2 as workaround to new compiler issues
- Added workaround for missing interfaces file on some builds
- Pinned onboard drivers to wlan-ap for access point

# Other
- Whitespace and syntax cleanup
- Replacement placeholders wrapped in __ __
- Contrib - mesh-cleanlinks.sh script to disconnect bad 802.11s links closed
- Better ExecStop command for all services
- Additional handling of non-critical errors during install
- Corrected iptunnel adding nat for each connection
- Firewall script update
- Added new NTP server

makeworld-the-better-one:
>NodeExporter reports mac addresses for mapping

Might this, and other static information, be better handled by nodeinfo.json? I thought nodeexporter was best for dynamic info like cpu usage and things.

darkdrgn2k:
it provides signal strength of mesh point links which are Mac mapped. 

they can be  mapped to cjdns peers (right now by guessing but new cjdns will allow it to be more accurate) 

but content is dynamic

FAQ.md
```diff
@@ -91,3 +91,14 @@ setenv load_script 'if test -e mmc 0:1 boot/boot.scr; then echo \"... booting fr
 setenv bootcmd 'run get_images; run set_bootargs; run load_script;booti \$kernel_addr \$ramfs_addr \$fdt_addr'
 saveenv
 ```
+**Q:** Seems all my mac addresses are the same across multiple boards. How do I fix this?
```
benhylau: Is this specific to Espressobin? If not, it should be in the Armbian section of this file.

contrib/ipfs-usb/movestore
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+if [ -z "$1" ] && [ -z "$2" ]; then
```
benhylau: What is `$1` and `$2`? Assign them to properly named variables at start of file.

FAQ.md
darkdrgn2k: Its an Arabian bug. not exclusive to espresso bin.

contrib/ipfs-usb/movestore
darkdrgn2k: There is a "syntax" section if you don't use them. But i added meaningful variables.

FAQ.md
benhylau: Shift up the doc in Armbian section :)

contrib/ipfs-usb/movestore
benhylau: Thanks

FAQ.md
darkdrgn2k: Duu :P :+1: lol

contrib/ipfs-usb/movestore
```diff
@@ -0,0 +1,20 @@
...
+user="$2"
+
+mkfs.ext3 /dev/$device
+service ipfs stop
+e2label  /dev/$device ipfs-data
+mkdir /mnt/ipfs-data
+echo "LABEL=ipfs-data  /mnt/ipfs-data ext3  defaults,noatime  0  0" >> /etc/fstab
+mount /mnt/ipfs-data
+mv /home/$user/.ipfs /mnt/ipfs-data/
+ln -s /mnt/ipfs-data/.ipfs /home/$user/.ipfs
```
makeworld-the-better-one: It might be better to just change the IPFS_PATH var, that way you don't have to remove symlinks and stuff later. Also it might be good to have a script to undo this.

scripts/ssb-web/uninstall
```diff
@@ -0,0 +1,21 @@
...
+
+sudo rm -rf /var/www/backend
+sudo rm -rf /var/www/sbot
+sudo rm -rf /usr/lib/node_modules
+sudo rm -rf /var/www/.ssb
+sudo rm -f /usr/local/bin/ssb-web-broadcast-service.sh
+
+sudo rm -rf /etc/nginx/site-path-enabled/ssb-web.conf
+
+sudo npm uninstall -g --unsafe-perm \
```
makeworld-the-better-one: Do we need sudo sudo here too? In case it's trying to remove root owned folders after dropping out of sudo? Or does that happen with post-**install** scripts only.

darkdrgn2k: so far its post-install. Well see if its a problem

contrib/ipfs-usb/movestore
darkdrgn2k: I'd like to revisit the IPFS_PATH in favor of a user instead anyway.

PR #188, opened 2018-08-28
Mostly patches for Debian OS

darkdrgn2k:


scripts/install2
```diff
@@ -88,6 +88,8 @@ fi
 
 echo -e "\e[1;32mStarting installation on ${BOARD_FAMILY} ${BOARD_NAME} (${BOARD_REVISION})...\e[0m"
 
+# Normalize OS environment
+
```
darkdrgn2k: The extra space here separates the SECTION comment from the next line comment

IE: This section is to normalize the OS  environment

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -33,4 +33,4 @@ while read -r cjdns_peer; do
 done <<< `sudo nodejs /opt/cjdns/tools/peerStats 2>/dev/null | awk '{ if ($2 == "ESTABLISHED") print $1 }' | awk -F. '{ print $6".k" }' | xargs`
 
 # Update peers data since ipfs just started
-/usr/local/bin/nodeinfo-update.sh
+sudo /usr/local/bin/nodeinfo-update.sh
```
makeworld-the-better-one: Does this mean that future nodeinfo modules will need to work under sudo? Like because this sudo will execute `run-parts` with sudo, which will execute all the nodeinfo files (just IPFS for now) with sudo. The IPFS script just curls from the 5001 api, but if it used just the `ipfs id` command then this sudo would have made it fail, if my thinking is correct. Would this be a problem in the future, where scripts might need to not use sudo?

darkdrgn2k: The other places that this runs are all run as root there are also some permission issues that's creep up if you don't run it as sudo although they can be mitigated with it running as root in other places anyway I figured this was the shortest path and somewhat standardized

darkdrgn2k: On issue i found is basically 
https://github.com/tomeshnet/prototype-cjdns-pi/blob/da224ec2f36d985c3de859829af8c21a0d74eafb/scripts/shared/nodeinfo/install#L12
Copies over a file with root write perms and
https://github.com/tomeshnet/prototype-cjdns-pi/blob/da224ec2f36d985c3de859829af8c21a0d74eafb/scripts/shared/nodeinfo/nodeinfo-update.sh#L3
Tries to write it but it cant unless its sudoed

scripts/install2
```diff
@@ -341,7 +352,6 @@ fi
 
 # Install possibly missing tools
 sudo apt-get install haveged -y || true
-sudo apt-get install iw -y || true
```
benhylau: Are you sure this isn't needed? It must have been added because one of the Armbians don't come with it... was it the Espressobin?

darkdrgn2k: its a revert to a commit done only a few days ago when we thought it didnt include it in Debian, turned out to be a path problem. Its just a correction

PR #191, opened 2018-08-29
Cjdns Verison bump

darkdrgn2k:
Version bump to a crashey version (CJD's input was queried)
New binaries created for this version and linked as well
Some scripts updated with new output of cjdns tools

- [ ] Test yrd @makeworld-the-better-one


scripts/install2
```diff
@@ -2,7 +2,7 @@
 
 set -e
 
-TAG_CJDNS=186169f9a8631633795e4e9d70c501519a7800f4
+TAG_CJDNS=d2e55d58548d83940482fe1bbbe1fd36f7f1b4ef
```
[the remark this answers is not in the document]
darkdrgn2k: I spoke to CJD and asked him what should i be using
He said use crashey as of today, but he reserves to change his statement tomorrow
so i pinned the commit

scripts/ipfs/ipfs-swarm-cjdns.sh
```diff
@@ -30,7 +30,7 @@ while read -r cjdns_peer; do
     # XXX: The below command hasn't been working -- so for now only 1-hop peers are checked
     #peers+=$(cjdnstool query getpeers $peer | sed -e '1d;$d' |awk -F. '{ print $6".k" }')
 
-done <<< `sudo nodejs /opt/cjdns/tools/peerStats 2>/dev/null | awk '{ if ($2 == "ESTABLISHED") print $1 }' | awk -F. '{ print $6".k" }' | xargs`
+done <<< `sudo nodejs /opt/cjdns/tools/peerStats 2>/dev/null | awk '{ if ($3 == "ESTABLISHED") print $2 }' | awk -F. '{ print $6".k" }' | xargs`
```
makeworld-the-better-one: Seeing things like this, I wonder how yrd will be affected. It should be tested before merging, it might not be valid anymore, or it might need a rebuild. It's last commit to master was on April 20th.

darkdrgn2k: It depends how yrd does things. If its direct to the api you may be fine as peerStats polls the api and the returns values.

I'm not familiar with yrd nor have any idea how to actually use it:)

benhylau: I wonder if cjdns itself will soon have a `cjdnsctl` type of program bundled in. In the meantime yrd is the best candidate. I think this is reasonable to make a core dependency. @darkdrgn2k thoughts?

darkdrgn2k: peerStats uses a api wrapper for node js

I know CJD has got some new command line tools working. For one he has a working traceroute finally :) Should get looked into.But its still node.js from what i understand

Only thing that i find odd about yrd is the service that needs to run, why would a "ifconfig" like app need a service?!?!

makeworld-the-better-one: >Only thing that i find odd about yrd is the service that needs to run, why would a "ifconfig" like app need a service?!?!

Yeah idk. I think it runs ok without it, :question:

PR #195, opened 2018-08-29
Install script refactor script

darkdrgn2k:


scripts/install2
```diff
@@ -2,6 +2,74 @@
...
+function checkModule {
+     eval res=\$$1
+     if [ ! -z "$res" -a "$res" == "true" ]; then
+        echo "1"
+    else
+        echo ""
+    fi
+}
+
 TAG_CJDNS=186169f9a8631633795e4e9d70c501519a7800f4
```
benhylau: I still think constants should be at the very top/

scripts/install2
```diff
@@ -2,6 +2,74 @@
...
+    fi
+
+    # This reads variable repersented by the string
+    eval res=\$$1
+
+    if [ "$(checkModule 'WITH_DIALOG')" ]; then
+
+        # Do not stop exec on non 0 return values
+        set +e
+        dialog $dialogGlobalParams $dialogParam --title "$2" --yesno "Install $2?" 6 55
```
benhylau: Can't do `|| true`?

scripts/install2
```diff
@@ -2,6 +2,74 @@
...
+        response=$?
+        # Return to previous setting
```
benhylau: Newline before comment.

scripts/install2
```diff
@@ -2,6 +2,74 @@
...
+        nonDefaultValue=true
+        defaultValue=false
+        dialogParam=" --defaultno "
+    fi
+
+    # This reads variable repersented by the string
+    eval res=\$$1
+
+    if [ "$(checkModule 'WITH_DIALOG')" ]; then
+
```
benhylau: Rm

scripts/install2
```diff
@@ -2,6 +2,74 @@
 
 set -e
 
+dialogGlobalParams="--backtitle Installation --ascii-lines"
+# Ask if module is to be enabled if not defined
+#    askModule <Variable> <Description> [default answer]
+function askModule {
+
```
benhylau: Rm

scripts/install2
```diff
@@ -119,213 +187,43 @@ if [ -z $(echo $PATH | grep "/sbin")  ]; then
...
-        echo ""
-        if [[ $REPLY =~ ^[Yy]$ ]]; then
-            echo -e "\e[1;32mSSB web client will be installed\e[0m"
-            WITH_SSB_WEB=true        
-        else
-            echo -e "\e[1;31mSSB web client installation will be skipped\e[0m"
-            WITH_SSB_WEB=false
-                fi
-    fi
+    askModule "WITH_IPFS_PI_STREAM" "IPFS Pi stream" "n"
```
benhylau: This seems like a good time to standardize our capitalization strategy. The messages have capitalizations that are kind of all over the place.

scripts/install2
darkdrgn2k: `set -e` stops executing whenever the return value is not 0 (it thinks there is an error) and in MOST cases this is the case. 0 for success 1 for failure. Sometimes other numbers .

By adding `|| true` overwrites the return code with 0 (since that's what true returns)

Dialog sets the return code to the selected item.  In the case of Yes/No  0 for Yes 1 for No.  This means that whenever some one selects No  `set -e` will stop the script.  And if we add  `|| true` it will simply overwrite all responses with Yes  (0)

Only solution i have so far is to `set +e` for the duration of the dialog and set it back to `set -e` right after.  If there is a better solution i would welcome it, or at very least detect if -e is set before applying it again after.

scripts/install2
```diff
@@ -2,6 +2,73 @@
...
+    # Define standard behaviour (default yes)
+    askPrompt="(Y/n)"
+    nonDefaultMatch="Nn"
+    defaultValue=true
+    nonDefaultValue=false
+    dialogParam=""
+
+    # Define alternative behaviour (default no)
+    if [ "$3" == "n" ]; then
+        askPrompt="(N/y)"
```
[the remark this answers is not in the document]
benhylau: Agree with `y/N` ordering. I have seen `()` and `[]` but think perhaps `[]` is a good change now that we use `()` for the component text itself such as `(experimental)`.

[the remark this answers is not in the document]
darkdrgn2k: Agread.

scripts/install2
darkdrgn2k: Decided to do Title Case (or whichever one it is that does not capitalize small words only)

scripts/install2
```diff
@@ -4,6 +4,73 @@ set -e
 
 TAG_CJDNS=186169f9a8631633795e4e9d70c501519a7800f4
 
+dialogGlobalParams="--backtitle Installation --ascii-lines"
+# Ask if module is to be enabled if not defined
+#    askModule <Variable> <Description> [default answer]
+function askModule {
+    # Define standard behaviour (default yes)
+    askPrompt="[Y/n)]"
```
makeworld-the-better-one: Extra paren.

scripts/install2
```diff
@@ -112,220 +179,50 @@ if [[ -e '/etc/NetworkManager/NetworkManager.conf' ]]; then
...
-        WITH_WATCHDOG=false
-    else
-        echo -e "\e[1;32mHardware Watchdog will be configured\e[0m"
-        WITH_WATCHDOG=true
-    fi
+askModule "WITH_H_DNS" "Hyperboria Domain Name Servers"
+askModule "WITH_H_NTP" "Hyperboria Network Time Protocol Server"
+askModule "WITH_EXTRA_TOOLS" "Non-Essential Tools Useful for Network Analysis"
+if [ "$SUPPORT_WATCHDOG" == "true" ]; then
+    askModule "WITH_WATCHDOG" "Hardware Watchdog"
```
makeworld-the-better-one: You mentioned to me that it should always be installed. Should it be an option at all? And if it should still, can the description here point out how it is required?

darkdrgn2k: Watchdog resets the cpu when it decides its broken. I'd rather give the user the choice if they want to hand over that power or not.

scripts/install2
darkdrgn2k: AAhhh..... thanks

scripts/install2
makeworld-the-better-one: Alright. It might be good to point people towards the readme in this repo at the beginning of the install though, as all the modules are explained there.

darkdrgn2k: We need to do a better job of illustrating to module in general.  I think a separate page with module descriptions and usage

darkdrgn2k:
I'd like to merge this as its a blocker for new modules.

Anything else that needs to be addressed or issues created

PR #205, opened 2018-09-04
default Install SSB web client (Experimental) (y/N)

NukeManDan:


darkdrgn2k:
@NukeManDan - Thank you for the submissions.  This issue was already resolved in the Development branch. 

I don't think we want to push anything into master (as it is the release tree). @benhylau @makeworld-the-better-one  thoughts:

NukeManDan:
When is the next release date? and what branch is the most up-to-date that I should use for testing ad-hock with some pi 3B and 3B+ ? https://github.com/tomeshnet/prototype-cjdns-pi/tree/develop ?

makeworld-the-better-one:
Master should be fine, you could also try develop (but reflash first), although develop shouldn't be used in production obviously. I would warn you that ad hoc tends to have bad drivers, if you can get your hands on other hardware you can try meshing over 802.11s (mesh point).

https://github.com/phillymesh/802.11s-adapters

darkdrgn2k:
> When is the next release date? 
No idea.  I have a few new features coming out in V4 but mostly going to be cleanup and refactor some of the bloat. I am hoping also to add a few more routing protocols to the stack,

>What  branch is the most up-to-date that I should use for testing 
Develop is the one we are actively working all PRs we do write against it.  I would use that (but expect some hickups here and there)

> ad-hock with some pi 3B and 3B+
I have NOT been able to get 3B to work in Ad-Hoc mode using the ONBOARD wifi. I have open tickets with Rapsberry Pi regarding this but it does not seem to help.  3B+ however seems to get into Ad-Hoc mode. If you want to try it check https://github.com/tomeshnet/prototype-cjdns-pi/blob/master/FAQ.md#wireless for how to do it 

> Master should be fine
Master is the latest release

> you could also try develop
Yes please feedback is good :+1: 

> but re flash first
Officially the stack ONLY woks if you re-flash each install.  Upgrades or re-installs are not supported official (although SOMETIMES may work, reinstall more so then upgrades)

>  develop shouldn't be used in production obviously. 
Its a prototype, and as far as i know there is no "production" network at all!

> I would warn you that ad hoc tends to have bad drivers
Yes many manufactures have legacy ad-hoc support and never check them. Strange things can happen to with some drivers. Like it says its doing X but in fact its doing Y.

> 802.11s (mesh point).
Most devices that support this the support is fairly new, and as much better chance of actually working.

darkdrgn2k:
PS: We also sit in this [chat](https://chat.tomesh.net/#/room/#tomesh:tomesh.net) all the time.
Bridged to FreeNode on #tomesh

if you want any help/support :)

PR #218, opened 2018-10-11
Corrected IPERF3 port number

darkdrgn2k:


darkdrgn2k:
That looks like there is duplication in that file... ill take a closer look!

Issue #220, opened 2018-10-14
Update firewall ports


darkdrgn2k:
Current firewall ports opened by default

# TCP IPv6
 - 22 - Shell
 - 80 - HTTP (ipfs gateway, patchfoo, ssb-web etc)
 - 443 - Nothing currently
 - 9100 - Node exporter
 - 9090 - Prometheus
 - 3000 - Grafana
 - 5201 - iperf3 tests
 - 4001 - IPFS swarm

#  UDP IPv4
 - ICMP
 - 67-68 -  DHCP
 - 53 - DNS
 - 8008 - ssb

# TCP IPv4
 - 22 SSH
 - 53 - DNS
 - 80  - HTTP (ipfs gateway, patchfoo, ssb-web etc)
 - 443 - Nothing currently 
 - 9100 - Node Exporter
 - 9090 - Prometheus
 - 3000 - Grafana
 - 5201 - iperf3
 - 4001 - IPFS Swarming

## Note
Prometheus and Grafana are currently only available on ipv4 (which usually is the WiFi interface)  

 - Do we open then up in ipv6 (hyperboria) or ask user if they would like to do that?
   - Added grafana to ipv6
 - Any other ports we need to open
 - Other comments concerns etc.

**EDIT**

Added 8008 port for ssb peering ipv4 udp
Added grafana and prometheus to open ports on ipv6

PR #221, opened 2018-10-15
Fixed rules

darkdrgn2k:
Issue #220 

Corrections made
Removed duplication of ipv6 rules
Removed unused of ipv6 CJDNS target

### DIFF
--- a/scripts/firewall/rules.v6
+++ b/scripts/firewall/rules.v6
@@ -6,6 +6,11 @@
 -A INPUT -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 -A INPUT -p ipv6-icmp -j ACCEPT
 -A INPUT -i lo -j ACCEPT
--A INPUT -j ACCEPT -p tcp -m multiport --dports 80,443,9100,22,5201,4001
+-A INPUT -j ACCEPT -p tcp --dport 22
+-A INPUT -j ACCEPT -p tcp --dport 80
+-A INPUT -j ACCEPT -p tcp --dport 443
+-A INPUT -j ACCEPT -p tcp --dport 5201
+-A INPUT -j ACCEPT -p tcp --dport 4001
+-A INPUT -j ACCEPT -p tcp --dport 9100
 -A INPUT -j REJECT --reject-with icmp6-port-unreachable
 COMMIT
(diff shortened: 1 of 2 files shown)


darkdrgn2k:
Note: Branched from master by accident, Travis is wrong on this one.

scripts/firewall/rules.v6
```diff
@@ -6,13 +6,6 @@
 -A INPUT -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 -A INPUT -p ipv6-icmp -j ACCEPT
 -A INPUT -i lo -j ACCEPT
--A INPUT -i tun0 -j cjdns
 -A INPUT -p tcp -m multiport --dports 80,443,9100,22,5201,4001
```
benhylau: Prometheus server and Grafana ports?

darkdrgn2k: C/P  issue #220

> Prometheus and Grafana are currently only available on ipv4 (which usually is the WiFi interface)

> Do we open then up in ipv6 (hyperboria) or ask user if they would like to do that?

I'll assume thats a yes?

benhylau: Why do we open some and not others? Where do we decide these are the ones we ask user about? If anything, it'd be node exporter that is sharing user information, not the prom server / grafana which just browse stuff right?

scripts/firewall/rules.v6
```diff
@@ -6,13 +6,6 @@
 -A INPUT -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
 -A INPUT -p ipv6-icmp -j ACCEPT
 -A INPUT -i lo -j ACCEPT
--A INPUT -i tun0 -j cjdns
--A INPUT -p tcp -m multiport --dports 80,443,9100,22,5201,4001
+-A INPUT -j ACCEPT -p tcp -m multiport --dports 80,443,9100,22,5201,4001
```
benhylau: Order the ports properly?

scripts/firewall/rules.v6
darkdrgn2k: Node Exporter - Point in time snapshot
Prometheus Server - Node Exporter data from any time that it was running
Grafana - Displays the above in a nice cpu intensive graph.

So one could argue that it gets progressively more invasive.  historical vs point in time.

Also grafana and prom are cpu/io hungry and cant really be used as a stack to monitor anything else but the local node really.

But I'm cool with opening those ports up if you still think it has value.

PR #222, opened 2018-10-15
Updated meshpoint script to scan all interfaces

darkdrgn2k:
Old mesh point scanned only phy0 and phy1, 

Update scans all interfaces, checks if they are wireless, determines the phy id, then checks for meshpoint. This way any phyx can be mesh point not just 0 and 1

### DIFF
--- a/scripts/mesh-point/mesh-point
+++ b/scripts/mesh-point/mesh-point
@@ -6,30 +6,35 @@ function isMeshable {
     # Check to see if driver reports missing mesh point support
     if  [ -z "$(iw phy phy$1 info | grep 'mesh point')" ]; then
         result='' # Indicate interface is not meshable
-    # XRADIO driver reports Mesh Point support incorrectly (there is no support for it)
-    elif [ "$(basename $(readlink /sys/class/net/$interface/device/driver))" == 'xradio_wlan' ]; then	    
+    # XRADIO driver reports Mesh Point but does not actually work
+    elif [ "$(basename $(readlink /sys/class/net/$interface/device/driver))" == 'xradio_wlan' ]; then
         result='' # Indicate interface is actually not meshable
     else
-        result='1' # Indicate interface is  meshable
+        result='1' # Indicate interface is meshable
     fi
 
     echo $result
 }
-
 
 set -e
 
 # Set wireless regulatory domain
 sudo iw reg set CA
 
-# Select physical device that supports 802.11s Mesh Point
-if ! [ -z $(isMeshable 0) ]; then
-    mesh_phy=phy0
-    mesh_dev=$(iw dev | grep phy#0 -A 1 | grep Interface | awk '{print $2}')
-elif ! [ -z $(isMeshable 1) ]; then
-    mesh_phy=phy1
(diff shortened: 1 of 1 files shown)


scripts/mesh-point/mesh-point
```diff
@@ -6,30 +6,35 @@ function isMeshable {
...
 
-# Select physical device that supports 802.11s Mesh Point
-if ! [ -z $(isMeshable 0) ]; then
-    mesh_phy=phy0
-    mesh_dev=$(iw dev | grep phy#0 -A 1 | grep Interface | awk '{print $2}')
-elif ! [ -z $(isMeshable 1) ]; then
-    mesh_phy=phy1
-    mesh_dev=$(iw dev | grep phy#1 -A 1 | grep Interface | awk '{print $2}')
-else
+# Find first meshpoint device you can
```
benhylau: It's better to use `802.11s Mesh Point` to be clear. The word meshpoint gets used in so many different context.

scripts/mesh-point/mesh-point
```diff
@@ -6,30 +6,35 @@ function isMeshable {
...
+for int in $(find /sys/class/net/* -maxdepth 1 -print0 | xargs -0 -l basename); do
```
benhylau: Would this not work if a phy interface is not already brought up in some other mode? It will only have a `phy` but not a `wlan` interface right?

darkdrgn2k: Good point.  
However when you plug in a wireless device it creates both a wlan? and a phy? regardless of being put into a mode (default mode is master).

This is not much different then the way it was done in v 2 and v 3 (and maybe v1) where we need a wlan to work, and never got any reports of issues.

We removed creating of any virtual interfaces (ie mesh0) because it broke stuff on raspbian and purpose the default one.

darkdrgn2k:
Also for when interface flap, currently the script wont re-run

Two thoughs i had

1) hook this into udev. But then we need a way to see if a meshpoint already exists

or 

2) Create a loop in the script to check for the created interface and if it goes "offline" fail the script so the service restarts

scripts/mesh-point/mesh-point
```diff
@@ -21,7 +21,7 @@ set -e
 # Set wireless regulatory domain
 sudo iw reg set CA
 
-# Find first meshpoint device you can
+# Find first 802.11s Mesh Point capable device you can
```
benhylau: Why does the sentence just break off? Isn't this a better sentence?

>`# Find first 802.11s Mesh Point capable device`

scripts/mesh-point/mesh-point
benhylau: Unless there is a way to do this logic within `phy`, I am fine with this, though not assuming virtual interfaces are created feels cleaner :)

benhylau:
Fix the comment before merge though.

scripts/mesh-point/mesh-point
darkdrgn2k: I agrea.

I like orange-mesh model for bringing these up personally but that will involve a bit more thinking/planning I think.  

Cause i dont think its wise to bring up every device in mesh mode?

Issue #198, opened 2018-09-01
Force password reset on pi


darkdrgn2k:
Force the pi user to reset the password on first login after install reboot? 

Expires pi users password
```
sudo passwd -e pi
```

jedahan:
First boot experience vs. security tradeoff.
If the expiration is presented as "Please choose a password!" then it empowers the user and improves security.
If the expiration also works fine via ssh / headless then 👍

darkdrgn2k:
armbian does this on first boot of the device on first login ssh or console

I think adding it to the rpi after first boot makes sense since it could be connected to the public hyperborea at that time

PR #224, opened 2018-11-13
Expire Password on Raspberry PI

darkdrgn2k:
Addresses issue #198

### DIFF
--- a/scripts/install2
+++ b/scripts/install2
@@ -86,7 +86,6 @@ SUPPORT_HOSTAP=false
 # Flag to identify if board can support on-board hardware watchdog.
 SUPPORT_WATCHDOG=false
 
-# Variables
 CJDNS_PACKAGE=""
 CHANGE_PASSWORD=""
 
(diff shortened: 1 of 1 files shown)


scripts/install2
```diff
@@ -421,6 +425,12 @@ newhostname=$(sudo grep -m 1 '"ipv6"' /etc/cjdroute.conf | awk '{ print $2 }' |
 sudo echo "$newhostname" | sudo tee /etc/hostname > /dev/null
 sudo sed -i -e "s/$oldhostname/$newhostname/" /etc/hosts
 
+if [[ "$CHANGE_PASSWORD" == "pi" ]]; then
+    sudo passwd -e "$CHANGE_PASSWORD"
```
benhylau: I didn't think sudo is required

scripts/install2
```diff
@@ -126,6 +129,7 @@ elif [[ "$BOARD_HARDWARE" == 'nanopineo2' || "$BOARD_HARDWARE" == 'sun50iw1p1' |
 else
     BOARD_FAMILY="Raspberry Pi"
     BOARD_REVISION="$(sed -rn 's/Revision\s+\:\s+([0-9a-z_\-\s\,\(\)]+)/\1/p' /proc/cpuinfo)"
+    CHANGE_PASSWORD="pi"
```
benhylau: There is forcing password change when running the script. Perhaps the user already changed the password? Can we detect for default password?

scripts/install2
darkdrgn2k: Seems only root can expire a password

```
pi@tomesh-6079:~ $  passwd -e pi
passwd: Permission denied.
pi@tomesh-6079:~ $ 
``

scripts/install2
darkdrgn2k: Hard coded default password hash and check it against `/etc/shadow`

Issue #219, opened 2018-10-14
cjdns-iptunnel does not work unless hostapd side effects occurred


benhylau:
To get cjdns-iptunnel working, we need to enable packet forwarding and NAT:

```
sudo sed -i 's/#net.ipv4.ip_forward=1/net.ipv4.ip_forward=1/' /etc/sysctl.conf
sudo sed -i 's/#net.ipv6.conf.all.forwarding=1/net.ipv6.conf.all.forwarding=1/' /etc/sysctl.conf

iptables -t nat -A POSTROUTING -o YOUR_INTERNET_IFACE -j MASQUERADE
```

These were done by installing hostapd on Raspberry Pis. In my case, it's a VM without an AP, but this also means if we install a Pi as an exit, without using the hostapd feature, the cjdns-iptunnel feature will not work.

I am thinking to add the first two lines to `scripts/cjdns-iptunnel/install` and the iptables line to `scripts/cjdns-iptunnel/cjdns-setup`. The first two lines are idempotent but the last one shouldn't be run if hostapd already did it right? Is there a better place to do these?

benhylau:
In the case of the VM, your fixes here is also needed https://github.com/tomeshnet/prototype-cjdns-pi/issues/175

darkdrgn2k:
I was thinking about how we put too much weight on what happens in hostapd. I think we should look at maybe enabling forwarding regardless since there are so many permutations of different modules that could use it. Option two is enable them in every module.

Natting would have happened at cjdns-setup but was dieing because of the -e

Ref: https://github.com/tomeshnet/prototype-cjdns-pi/blob/d73e61268ab06825c13eb19d1ccaec3bce69072a/scripts/cjdns-iptunnel/cjdns-setup#L38

You may want to add a sleep at the top cause there is a bit of a race condition where cjdns may not have the tun0 up by the time you are ready.

However we may be able to also get away with using tun+ instead of tun0 to identify any tun interface.

darkdrgn2k:
Corrected in cjdns-setup

https://github.com/tomeshnet/prototype-cjdns-pi/blob/cjdns-iptunnel-ipv6/scripts/cjdns-iptunnel/cjdns-setup#L9

benhylau:
I agree with all you said, and this option:
>enabling forwarding regardless

PR #227, opened 2018-11-22
Cjdns tunnel fix

### LINKED ISSUE #175: IPTunnel - nodejs does not run under udev on some computers
Possible issues on some devices running nodejs (cexec) via udev

https://github.com/nodejs/node/issues/19036

https://github.com/nodejs/node/issues/16901

darkdrgn2k:
#175 

Swap out udev cjdns-setup with execstartpost cjdns-setup

### DIFF
--- a/scripts/cjdns-iptunnel/cjdns-setup
+++ b/scripts/cjdns-iptunnel/cjdns-setup
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+sleep 8
 
 SUBNET4="10.1.0."
 SUBNET6="fe80::"
--- a/scripts/cjdns-iptunnel/cjdns-setup
+++ b/scripts/cjdns-iptunnel/cjdns-setup
@@ -1,6 +1,8 @@
 #!/usr/bin/env bash
 
-sleep 8
+# The service should have started by now and tun interface should be up
+# But if not we can uncomment this sleep line
+# sleep 8
 
 SUBNET4="10.1.0."
 SUBNET6="fe80::"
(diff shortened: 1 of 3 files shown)


scripts/cjdns-iptunnel/cjdns-setup
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e
+sleep 8
```
benhylau: I don't think this fix was necessary in my case. Can we check if these two steps are sufficient?
https://github.com/tomeshnet/prototype-cjdns-pi/issues/219#issuecomment-429618886

Really don't want a `sleep <arb value>` unless we must. And if we really must then this needs a comment to explain.

scripts/cjdns-iptunnel/install
```diff
@@ -5,5 +5,16 @@ set -e
...
 # Install scripts that run after cjdns tun0 is configured
-sudo cp "$BASE_DIR/50-cjdns.rules" /etc/udev/rules.d/50-cjdns.rules
 sudo cp "$BASE_DIR/cjdns-setup" /usr/local/sbin/cjdns-setup
+
+# Update service to start script on cjdns start
+if [ -f /lib/systemd/system/cjdns.service ]; then
+    sudo sed /ExecStartPost/d /lib/systemd/system/cjdns.service
+    sed -i s#Restart=always#Restart=always\\nExecStartPost=/usr/local/sbin/cjdns-setup# /lib/systemd/system/cjdns.service
+fi
+if [ -f /etc/systemd/system/cjdns.service ]; then
```
benhylau: Why do we have this in two different dir in the fist place?? Can we fix the real problem?

scripts/cjdns-iptunnel/cjdns-setup
darkdrgn2k: Im doing its "just in case" because we need tun0 to be up before proceeding.

i can drop it but.... its a precaution

scripts/cjdns-iptunnel/install
darkdrgn2k: @benhylau 
Its not so much as problem but how the install happend

`deb` packages install in `/lib/systemd/system`  because its "part of the system"
`compiled` installs in /etc/systemd/system` because its "user created" script

So it depends if you installed the deb or compiled.

darkdrgn2k: Made issue to do this #262

PR #228, opened 2018-11-22
Add HLS Player for IPFS Live Stream Repo

darkdrgn2k:
Added hls player into prototype to allow easy playing of ipfs stream.


README.md
```diff
@@ -56,7 +56,7 @@ Many board that run [Armbian](https://www.armbian.com/) such as many models of O
...
 | `WITH_IPFS`                     | **80**: HTTP-to-IPFS gateway at `/ipfs/HASH`   | Set to `true` if you want to install [IPFS](https://ipfs.io). |
-| `WITH_IPFS_PI_STREAM`           | None                                           | Set to `true` if you want to install Pi stream service to live stream your camera over IPFS. Requires a Raspberry Pi with camera module. *Will automatically start recording on boot by default.* |
+| `WITH_IPFS_PI_STREAM`           | None                                           | Set to `true` if you want to install Pi stream service to live stream your camera over IPFS. Requires a Raspberry Pi with camera module.  Player interface at `/video-player` *Will automatically start recording on boot by default.* |
```
benhylau: The sentence now reads very awkward... I think you mean...

>Automatically start recording on boot, playback available at `/video-player`.

One space after periods like everywhere else in this repo.

scripts/ipfs-pi-stream/install
```diff
@@ -24,3 +24,14 @@ sudo systemctl enable process-stream
 
 # Add hourly job to clear out old data
 echo "41 * * * * $USER /usr/local/bin/ipfs repo gc" | sudo tee --append /etc/crontab
+
+# Install the ipfs video player
+mkdir "$BASE_DIR/tmp"
+git clone https://github.com/tomeshnet/ipfs-live-streaming.git "$BASE_DIR/tmp/live-streaming"
```
benhylau: We better call this directory `ipfs-live-streaming`. In general, I prefer to avoid unnecessary name mapping in code.

scripts/ipfs-pi-stream/install
```diff
@@ -24,3 +24,14 @@ sudo systemctl enable process-stream
...
+cp -r "$BASE_DIR/tmp/live-streaming/terraform/shared/video-player" /var/www/html/video-player
```
benhylau: Do the `sed` in the `tmp` and copy at the end. This way we never get intermediate states in case something fails. If there is a file in `/var/www/html/video-player` we know the `sed` lines succeeded. This is also generally applicable.

scripts/ipfs-pi-stream/install
```diff
@@ -24,3 +24,14 @@ sudo systemctl enable process-stream
...
+sed -i s#__IPFS_GATEWAY_SELF__#/ipfs/# /var/www/html/video-player/js/common.js
+sed -i s#__IPFS_GATEWAY_ORIGIN__#https://ipfs.io/ipfs/# /var/www/html/video-player/js/common.js
```
benhylau: I don't think origin would be ipfs.io here, it's localhost? And it shouldn't include the `/ipfs/`

darkdrgn2k: M3U8 file is still created with ipfs.io as the path.  This way it can be used "as is"

The script does a replace of `__IPFS_GATEWAY_ORIGIN__` to `__IPFS_GATEWAY_SELF__` so it needs to know what to replace.

scripts/ipfs-pi-stream/install
```diff
@@ -24,3 +24,15 @@ sudo systemctl enable process-stream
...
 echo "41 * * * * $USER /usr/local/bin/ipfs repo gc" | sudo tee --append /etc/crontab
+
+# Install the ipfs video player
+mkdir "$BASE_DIR/tmp"
+git clone https://github.com/tomeshnet/ipfs-live-streaming.git "$BASE_DIR/tmp/ipfs-live-streaming"
+VIDEO_PLAYER_PATH="$BASE_DIR/tmp/live-streaming/terraform/shared/video-player"
+sed -i s#__IPFS_GATEWAY_SELF__#/ipfs/# "$VIDEO_PLAYER_PATH/js/common.js"
+sed -i s#__IPFS_GATEWAY_ORIGIN__#https://ipfs.io/ipfs/# "$VIDEO_PLAYER_PATH/js/common.js"
+IPFS_ID=$(ipfs id | grep ID | head -n 1 | awk -F\" '{print $4}')
+sed -i "s#live.m3u8#/ipns/$IPFS_ID#" "$VIDEO_PLAYER_PATH/js/common.js"
```
benhylau: Shall we change https://github.com/tomeshnet/ipfs-live-streaming/blob/master/terraform/shared/video-player/js/common.js#L6 instead of messing up the previous line?

scripts/ipfs-pi-stream/install
benhylau: Ah I didn't realize the script creates with ipfs.io I thought it would just make localhost by default

scripts/ipfs-pi-stream/install
darkdrgn2k: Agreaed,

however because of  __IPFS_GATEWAY_ORIGIN__  (which is already replaced) and just the amount of lines it would take to do that i think its "safer" to keep as is.

Added pin repo instead.

README.md
darkdrgn2k: re-worded as

> Set to true if you want to install Pi stream service to live stream your camera over IPFS. Requires a Raspberry Pi with camera module. Player interface at /video-player Warning: By default the camera will start broadcasting after first reboot.

Issue #210, opened 2018-09-16
Captive portal (WiFi discoverability for newbie users)


DaniellMesquita:
When connecting to WiFi, it shows a captive portal, where user can signup/login, and owner can customize it.

makeworld-the-better-one:
>where user can signup/login

Signup or login to what? Some services on the pi may eventually have a login service online, but having a captive portal for the whole pi doesn't make sense. As for allowing wifi access, a WPA password should be enough. EAP is also supported, which allows logins, although we don't have a portal for that at the moment. Feel free to add it for your own use though!

DaniellMesquita:
Not good for offline users to can't login WiFi because it haves no password.
Also, when they log on open WiFi, they doesn't knows how to join CJDNS.

DaniellMesquita:
@makeworld-the-better-one
Do you have a contact other than GitHub?

benhylau:
I also believe a captive portal is out of scope for this project. It is not intended to be a public access point allowing users to sign up new accounts. It's an installer to set up your own mesh node with cjdns and peer to peer apps.

bonfus:
What about a third option in `hostapd/install` which provides an open wifi with restrictions on private network addresses (i.e. 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, fd00::/8) ?
I'm drafted it on my installation and would be happy to open a PR if you think it fits the scope of your project.

makeworld-the-better-one:
>Not good for offline users to can't login WiFi because it haves no password. Also, when they log on open WiFi, they doesn't knows how to join CJDNS.

@

makeworld-the-better-one:
>Not good for offline users to can't login WiFi because it haves no password. Also, when they log on open WiFi, they doesn't knows how to join CJDNS.

@DaniellMesquita

Joining CJDNS isn't really part of the code, it only provides *access* to CJDNS. This happens automatically so you don't need to worry about users. 

If I understand correctly, your other concern is that if your WiFi has no password, users won't be able to have a login. I would reccomend a password for security, but if you don't have one, then everyone can still use your WiFi fine. The issue for you is with creating an account I think, which I am confused about. Why?

DaniellMesquita:
@bonfus

> What about a third option in hostapd/install which provides an open wifi with restrictions on private network addresses (i.e. 10.0.0.0/8, 172.16.0.0/12, 192.168.0.0/16, fd00::/8) ?
[...]

PR #229, opened 2018-11-23
[Contrib] Captive Portal like lock for internet bound connection on an offline node

darkdrgn2k:
Redirects to Pi's webserver when using WIFI Access Point on a node that will not route ipv4 (Internet) traffic.

- Removed IPV4 masquerading
- Addes redirect to 10.0.0.1 for all port 80 bound traffic
- Adds 404 redirect to / for nginx

### DIFF
--- a/contrib/offline/redirect-port-80.sh
+++ b/contrib/offline/redirect-port-80.sh
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+# Backup file
+if ! [ -f "/etc/hostapd/nat.sh.org" ]; then
+  sudo cp /etc/hostapd/nat.sh /etc/hostapd/nat.sh.org
+fi
+
 # Redirect all IPv4 80 traffic to the pi
 echo iptables -t nat -I PREROUTING -i wlan-ap -p tcp --dport 80 -j DNAT --to-destination 10.0.0.1:80 | sudo tee --append  /etc/hostapd/nat.sh > /dev/null 
 
@@ -9,3 +14,4 @@ sudo sed -i "/iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE/d" /etc/hosta
 
 # Set nginx to redirect any 404 errors to /
 sudo sed -i "s/}/    error_page 404 =200 \/index.html;\\n}/" /etc/nginx/sites-enabled/main.conf
+sudo systemctl restart hostapd
--- a/contrib/offline/redirect-port-80.sh
+++ b/contrib/offline/redirect-port-80.sh
@@ -1,17 +1,17 @@
 #!/bin/sh
 
 # Backup file
-if ! [ -f "/etc/hostapd/nat.sh.org" ]; then
-  sudo cp /etc/hostapd/nat.sh /etc/hostapd/nat.sh.org
+if ! [ -f "/etc/hostapd/nat.sh.bak" ]; then
+  sudo cp /etc/hostapd/nat.sh /etc/hostapd/nat.sh.bak
 fi
 
 # Redirect all IPv4 80 traffic to the pi
 echo iptables -t nat -I PREROUTING -i wlan-ap -p tcp --dport 80 -j DNAT --to-destination 10.0.0.1:80 | sudo tee --append  /etc/hostapd/nat.sh > /dev/null 
 
-# Prevent masquarading out ipv4
-# This is prevent IPTUNNEL and routing to the internet (Exit node)
+# Prevent masquerading out IPv4
+# This is to prevent IPTUNNEL and routing to the internet (Exit node)
 sudo sed -i "/iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE/d" /etc/hostapd/nat.sh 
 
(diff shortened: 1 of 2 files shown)


darkdrgn2k:
issue #210 

@DaniellMesquita - Thoughts?

darkdrgn2k:
>  Including an uninstall script,

Its a contrib file. Does it need an uninstall script? 
I made a backup of the nat file though.

contrib/offline/redirect-port-80.sh
```diff
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+# Backup file
+if ! [ -f "/etc/hostapd/nat.sh.org" ]; then
+  sudo cp /etc/hostapd/nat.sh /etc/hostapd/nat.sh.org
```
benhylau: I think in other places in the codebase we have been using `.bak` for backup files

contrib/offline/redirect-port-80.sh
```diff
@@ -0,0 +1,17 @@
...
+
+# Backup file
+if ! [ -f "/etc/hostapd/nat.sh.org" ]; then
+  sudo cp /etc/hostapd/nat.sh /etc/hostapd/nat.sh.org
+fi
+
+# Redirect all IPv4 80 traffic to the pi
+echo iptables -t nat -I PREROUTING -i wlan-ap -p tcp --dport 80 -j DNAT --to-destination 10.0.0.1:80 | sudo tee --append  /etc/hostapd/nat.sh > /dev/null 
+
+# Prevent masquarading out ipv4
```
benhylau: `masquerading`

benhylau: `IPv4` as above

contrib/offline/redirect-port-80.sh
```diff
@@ -0,0 +1,17 @@
...
+# This is prevent IPTUNNEL and routing to the internet (Exit node)
```
benhylau: `to` prevent

contrib/offline/redirect-port-80.sh
```diff
@@ -0,0 +1,17 @@
...
+sudo sed -i "/iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE/d" /etc/hostapd/nat.sh 
+
+# Set nginx to redirect any 404 errors to /
+sudo sed -i "s/}/    error_page 404 =200 \/index.html;\\n}/" /etc/nginx/sites-enabled/main.conf
```
benhylau: This replaces `}`? Is it safe?

benhylau:
@darkdrgn2k @makeworld-the-better-one contribs generally don't have uninstalls

contrib/offline/redirect-port-80.sh
darkdrgn2k: hacky I know
in this. file there should be only one}

don't know how else to do this Esp in bash

@benhylau any ideas? how to insert a line on the Second last line of a file

benhylau: `sed -i '$i    error_page 404 =200 /index.html;' /etc/nginx/sites-enabled/main.conf`

DaniellMesquita:
Basically explaining for dumbs, how it works?

darkdrgn2k:
It basically has 2 lines

### Firewall
`iptables -t nat -I PREROUTING -i wlan-ap -p tcp --dport 80 -j DNAT --to-destination 10.0.0.1:80 `

This means ANY IPv4 packets that will be forwarded to another device, destined for TCP port 80, re-write its destination to the node itself  10.0.0.1:80  (Break down of the command at the bottom if your interested)

### nginx
`error_page 404 =200 /index.html`

Because our nginx server PROBABLY doesn't have the random path the person is asking for (because we are not the server the person wanted to access) they would get a 404 error (IE http://99.99.99.99/myfile.html  instead  you will get the content of http://10.0.0.1/myfile.html and that file does not exist)

This command takes all 404 errors and instead presents them with /index.html (home page)


# Break down of iptables command

`iptables`
Firewall for IPv4

`-t nat -I PREROUTING`
We are using the `NAT` module. Rules go in the `PREROUTING` table. Processing of this table happens when a packet is do to forwarded to another device but before it is routed.

`-i wlan-ap`
Packet coming in on the Access Point

`-p tcp`
TCP Protocol

`--dport 80`
Destination Port 80 (TCP port 80 is HTTP)

`-j DNAT`
Tells the firewall we will be doing a Destination NAT.  This means that we will be changing the destination of the packet. 

`--to-destination 10.0.0.1:80`
Re write the packet destination to the wlan-ap ip address

Then the packet goes through the routing process and ends up on the node's web server instead of being sent of into the wild.

DaniellMesquita:
Then, it only accepts IPv6 traffic?

DaniellMesquita:
How this script can be ported to Android?

makeworld-the-better-one:
@DaniellMesquita IPv6 traffic is accepted, this way Yggdrasil and CJDNS addresses are not redirected. Porting to Android is far beyond the scope of this project, and I have no Android networking knowledge.

PR #252, opened 2018-12-13
Make nginx install by default

### LINKED ISSUE #248: Nginx installation
~~Nginx right now installes for IPFS and some SSB modules~~

~~Should it always be installed? It serves.~~
~~ - Welcome page~~
~~ - nodeinfo.json~~

~~also~~

~~- [ ] Add check if its installed before trying to write to it~~

- [x] Remove shared installation of nginx from modules
- [x] Move nginx from shared folder to main modules 
- [x] Install nginx as part of install script (unasked module)

makeworld-the-better-one:
Fixes #248. Yay!


scripts/install2
```diff
@@ -331,6 +331,9 @@ sudo systemctl enable cjdns.service
 NEWHOSTNAME=$(sudo grep -m 1 '"ipv6"' /etc/cjdroute.conf | awk '{ print $2 }' | sed 's/[",]//g' | sed "s/.*:/$MESH_NAME-/g")
 export NEWHOSTNAME
 
+# Install nginx without asking
```
benhylau: There is no need to emphasize `without asking`. The comment should just read `# Install nginx` not what it does not do.

scripts/ipfs/install
```diff
@@ -23,8 +23,8 @@ rm -rf "$BASE_DIR/tmp"
 ipfs init || true
 
 # Configure HTTP to IPFS gateway
-# shellcheck source=../shared/nginx/install
-source "$BASE_DIR/../shared/nginx/install"
```
benhylau: I don't understand why it is moved out of `shared`. The fact that it is installed by default doesn't change the fact that it's shared?

This also makes me think whether it makes sense to start sorting modules into `default` vs. `optional`, since every module folder so far is an optional component. Now `nginx` is default, like `cjdns`, which never had its own module and kind of live within the main install scripts. My concern is, the knowledge of which ones are default and which are not currently is not expressed in code, making this repo hard to maintain.

makeworld-the-better-one: I did it originally because @darkdrgn2k suggested it in #248. I agree that default and optional would be good categories to separate things into, although I don't really like the idea of `nginx` (and `cjdns`) not having their own module. At the moment there is only one module in `shared`, `nodeinfo`, so I would be in favour of moving that out as well. What if I do that, and then I can make a separate PR for the `default` and `optional` folders?

scripts/install2
makeworld-the-better-one: Done.

scripts/ipfs/install
darkdrgn2k: Nginx was previous installed only when needed (ipfs, patchfoo etc)
We have decided in that issue to make it a core because of nodeinfo.json and welcome page which we want to exist regardless of what is installed.

Therefor they now are part of the main script not a "default" install but a "required" install.

as for  CJDNS not being a module it is currently  tied in pretty hard into things (ie Access Point name and Host name and such) that would require more work to untangle.

scripts/ipfs/install
```diff
@@ -23,8 +23,8 @@ rm -rf "$BASE_DIR/tmp"
...
+# shellcheck source=../nginx/install
+source "$BASE_DIR/../nginx/install"
```
darkdrgn2k: Should no longer be installed as part of the module remove line 26, 27

already dont in main script
https://github.com/tomeshnet/prototype-cjdns-pi/pull/252/files#diff-70643e83d82dc35762cb692e2eb6e37cR335

scripts/ssb-patchfoo/install
```diff
@@ -7,8 +7,8 @@ BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 CURRENT_DIR="$(pwd)"
 
 # Install nginx
-# shellcheck source=../shared/nginx/install
-source "$BASE_DIR/../shared/nginx/install"
+# shellcheck source=../nginx/install
+source "$BASE_DIR/../nginx/install"
```
darkdrgn2k: Should no longer be installed as part of the module remove line 10,11

scripts/ssb-web-pi/install
```diff
@@ -7,8 +7,8 @@ set -e
 BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 # Install nginx and php
-# shellcheck source=../shared/nginx/install
-source "$BASE_DIR/../shared/nginx/install"
+# shellcheck source=../nginx/install
+source "$BASE_DIR/../nginx/install"
```
darkdrgn2k: Should no longer be installed as part of the module remove line 10,11

scripts/ipfs/install
makeworld-the-better-one: Of course, all these installs. I'll fix that.

benhylau:
This repository now has more than one/two developer. It would be helpful to start assigning issues to yourself _before_ working on it to prevent duplication.

makeworld-the-better-one:
No one was assigned to the issue at the time, but obviously I should've assigned it to myself after starting to work on it. I will next time.

darkdrgn2k:
Also my fault, github/my keyboard ate some letters when i was making the branch and it was spelled wrong.
Dint fix it cause i didn't think anyone else would want to deal with this.

@makeworld-the-better-one thanks for dealing with this :)

darkdrgn2k:
Fixed conflict from SDR pr

benhylau:
Yea we never did have a process of assigning issues. Seems like a good idea to start now :)

scripts/ipfs/install
benhylau: So which are we doing?
- nginx in `shared`
- classify into `default` vs. `optional`
- Leave nginx in root like cjdns

makeworld-the-better-one:
@benhylau @darkdrgn2k I'm thinking we move all modules into either `default` or `optional` folders. I can do that tomorrow if we agree. What do you think?

scripts/ipfs/install
darkdrgn2k: moved conversation to https://github.com/tomeshnet/prototype-cjdns-pi/issues/248

darkdrgn2k:
Moved conversation to issue, i want to merge some of these PRs in so i can work on the next round of updates to move to a release.

I just ran and tested this stack and it seem to work.

Issue #256, opened 2019-01-04
IPFS update to 0.4.18 (2018-10-26)


makeworld-the-better-one:
I'm making this issue about an IPFS update for discussion, but also as reminder to myself, or anyone else that wants to handle this PR before me.

> This is probably one of the largest go-ipfs releases in recent history, 3 months in the making.

The two major changes for us are a [new pubsub routing algorithm](https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#pubsub), and a nice [new WebUI](https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#webui). Further changes can be read about in the [changelog](https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#0418-2018-10-26). The decision I see us having to make is whether to enable the new pubsub routing algorithm, [gossipsub](https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md#gossipsub), as opposed to the current one called floodsub. Here's some quotes on the new algorithm:

> Gossipsub is a new, experimental routing protocol for pubsub that should waste less bandwidth than floodsub, the current pubsub protocol. It's backwards compatible with floodsub so enabling this feature shouldn't break compatibility with existing IPFS nodes.

> The gossipsub routing algorithm is significantly more efficient than the current floodsub routing algorithm.

> You can enable gossipsub via configuration: `ipfs config Pubsub.Router gossipsub`

Should we enable it? I know pubsub is used in some projects, namely live-streaming over IPFS. The only qualm about enabling would be that it is still experimental. Testing needed? Either way, an update to IPFS should happen soon. I'll assign myself this issue once I start, but if anyone else wants to grab it it's yours.

PR #261, opened 2019-01-23
IPFS version bump

darkdrgn2k:
Resolve issue #256


makeworld-the-better-one:
Why don't we also enable gossipsub?

PR #270, opened 2019-01-30
Initial modules example [ygg]

darkdrgn2k:


docs/MODULES.md
```diff
@@ -32,9 +32,32 @@ To install all optional modules (not recommended), run the following command:
 $ wget https://raw.githubusercontent.com/tomeshnet/prototype-cjdns-pi/master/scripts/install && chmod +x install && WITH_MESH_POINT=true WITH_AD_HOC=false WITH_WIFI_AP=true WITH_FIREWALL=true WITH_CJDNS_IPTUNNEL=true WITH_IPFS=true WITH_SSB=true WITH_SSB_WEB_PI=true WITH_PROMETHEUS_NODE_EXPORTER=true WITH_PROMETHEUS_SERVER=true WITH_GRAFANA=true WITH_H_DNS=true WITH_H_NTP=true WITH_EXTRA_TOOLS=true WITH_WATCHDOG=true WITH_YRD=true ./install
 ```
 
+## CJDNS
+Cjdns (Caleb James DeLisle's Network Suite) is a networking protocol and reference implementation. It is founded on the ideology that networks should be easy to set up, protocols should scale smoothly, and security should be built in by default.
+
+CJDNS uses cryptography to self-assign IPv6 address in the fc00/8 subnet and will automatically peer with other nodes connected via Layer2 ethernet, broadcasts or configured UDP tunnels.
```
makeworld-the-better-one: It should be `fc00::/8`.

docs/MODULES.md
```diff
@@ -32,9 +32,32 @@ To install all optional modules (not recommended), run the following command:
...
+
+For more information please see the [CJDNS FAQ](https://github.com/cjdelisle/cjdns/blob/master/doc/faq/general.md)
+
+### CJDNS Firewall
```
makeworld-the-better-one: These open ports also apply to Yggdrasil and any other interface. They are also mentioned above in the table of modules. If you want to list them again it might be better as its own section.

makeworld-the-better-one: Oh ok so I see you have done that already. I wouldn't list it here then.

makeworld-the-better-one: We will leave this for now, because of #272.

Issue #245, opened 2018-12-01
nodeinfo.json Yggdrasil support


makeworld-the-better-one:
The nodeinfo file on nodes should also provide information on yggdrasil: node id, keys, etc.

makeworld-the-better-one:
Yggdrasil currently doesn't have a spec for nodeinfo. The CJDNS spec is [here](https://github.com/cjdelisle/cjdns/blob/master/doc/cjdns/nodeinfo-json.md), and we need to figure out a way to combine Ygg info with that already existing spec. Once that is worked out it can be submitted to the [yggdrasil-specs](https://github.com/yggdrasil-network/yggdrasil-specs) repo.

makeworld-the-better-one:
Whatever is worked out here, possibly with @neilalexander's help, should be served over http and possibly through the Yggdrasil config file as well. At the moment, we have a `name` value setup, but that will be sort of unclear when put into the http nodeinfo, and so maybe we can change that later.

makeworld-the-better-one:
Whatever is worked out here, possibly with @neilalexander's help, should be served over http and possibly through the Yggdrasil config file as well. At the moment, we have a `name` value setup, but that will be sort of unclear when put into the http nodeinfo, and so maybe we can change that later.

neilalexander:
So as you identified we use `"name"` instead of `"hostname"`. We also have built-ins that are different from `"software"` in the form of `"buildarch"`, `"builldname"`, `"buildplatform"` and `"buildversion"`:
```
{
  "nodeinfo": {
    "buildarch": "amd64",
    "buildname": "yggdrasil-develop",
    "buildplatform": "darwin",
    "buildversion": "0.3.2-0080",
    "name": "imac.y.neilalexander.eu"
  }
}
```

I don't particularly like the `"services"` part either because the URI bit is either difficult to specify or inapplicable for many protocols. Something closer to [DNS-SD (RFC 6763)](https://tools.ietf.org/html/rfc6763) would be preferable, e.g.
```
"somehttpservice": {"_http", "_tcp", ttl, class, priority, weight, "1.2.3.4", 80}
```
... or similar, given that it makes it possible to either broadcast mDNS packets locally for those services or to populate SRV records in DNS zones.

I don't particularly like that the `"pgp"` section makes the assumption that a node will also have Internet access to reach SKS, because we have no guarantee of that, and also SKS is extremely broken anyway. 

[...]

PR #271, opened 2019-01-31
Add node's hostname to yggdrasil

darkdrgn2k:
This will label the node on the official yggdrasil map


darkdrgn2k:
> I'd add that this is only a temporary solution

This has nothing to do with #245  

This is a Yggdrasil native method of setting a name of the node to be accessible via yggdrasil.

Issue #263, opened 2019-01-25
Switch to 20Mhz band


darkdrgn2k:
Discussion: 
https://github.com/tomeshnet/prototype-cjdns-pi/pull/217#issuecomment-443344199

PR #273, opened 2019-01-31
20hmz band

darkdrgn2k:
Issue #263

Discussion welcome

----
CP from #217 
-----

### 20MHz 
- Less Noise = More stable
- Less throughput
- More supported devices

### 40Mhz
- More throughput available
- More noise - Possibly weaker link

### Issues
- If we switch mesh point old versions will not mesh with new versions.

# Longer description

2.4 GHz Wifi is made up of 11 channels spaced 5Mhz apart ~(total spectrum 50 MHz)

Because we use define channels by "Center frequencies" (that is the frequency that is in the middle of the band) There is another ~10MHz on each end of the spectrum. (As an example Channel 1 is known as 2412Mhz, because that's the middle.  It actually starts at 2401Mhz. ) That brings the total spectrum up to ~ 65 MHz.

Standard WIFI channel is 20MHz, this means it covers about 4-5 channels. (basically 2 channels in each direction of the selected channel)  Any overlapping channels make NOISE for the channels they overlap.
image

In a "Best case" scenario you dont want channels to be overlapping. This is why there they say to run wifi only on 1 6 11 because then they dont overlap. If they overlap then its just noise. This means there are only 3 non-overlapping usable channels on 2.4 Ghz. (65 Mhz /20Mhz=~3 + a bit of space for separating channels to prevent noise.)

40MHz is a little more difficult.  First you can only fit ONE 40mhz channel in 2.4 Ghz without overlapping.
image

## Current model
Current model we use 1 x 20MHz channel for Access point (channel 1) and 1 x 40Mhz channel for Mesh Point (and as such AdHoc etc)

- This takes over the whole spectrum.  
- Any other devices on channels 2, 3 running 20MHz create nose for the access point
- Any other devices on channels 1,2,3,4,5,6,7,8,9 running 40Mhz create nose for the access point
- Any devices running on channels 2,3,4,5,6,7,8,9,10,11 running 20MHz create noise for the mesh
- Any devices running on channels 2,3,4,5,6,7,8,9,10 running 40Mhz create noise for the mesh.

/_\ if there is allot of devices on any other channels/bands that we use we get allot more noise.

### DIFF
--- a/scripts/mesh-adhoc/mesh-adhoc
+++ b/scripts/mesh-adhoc/mesh-adhoc
@@ -20,9 +20,9 @@ sudo ifconfig $mesh_dev up
 # Optionally assign IPv4 address to the mesh_dev interface
 # sudo ifconfig $mesh_dev 192.168.X.Y
 
-# Join the mesh network with radio in HT40+ htmode to enable 802.11n rates
-##TODO## Check for HT40+ first
-sudo iw dev $mesh_dev ibss join MESH_NAME 2412 HT40+
+# Join the mesh network with.
+# To join radio in HT40+ htmode (enable 802.11n rates) add  HT40+ to end of this line
+sudo iw dev $mesh_dev ibss join MESH_NAME 2412
 
 # Restart cjdns
 sudo killall cjdroute
--- a/scripts/mesh-adhoc/mesh-adhoc
+++ b/scripts/mesh-adhoc/mesh-adhoc
@@ -20,7 +20,7 @@ sudo ifconfig $mesh_dev up
 # Optionally assign IPv4 address to the mesh_dev interface
 # sudo ifconfig $mesh_dev 192.168.X.Y
 
-# Join the mesh network with.
+# Join the mesh network
 # To join radio in HT40+ htmode (enable 802.11n rates) add  HT40+ to end of this line
 sudo iw dev $mesh_dev ibss join MESH_NAME 2412
 
--- a/scripts/mesh-point/mesh-point
+++ b/scripts/mesh-point/mesh-point
@@ -50,8 +50,9 @@ sudo ifconfig $mesh_dev up
 # Optionally assign IPv4 address to the mesh_dev interface
 # sudo ifconfig $mesh_dev 192.168.X.Y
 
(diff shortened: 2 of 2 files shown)


makeworld-the-better-one:
> Will 20mhz actually have a noticeable difference for the applications we use and testing we do?

My only question before merging. I'm guessing the answer is no.

darkdrgn2k:
> 
> 
> > Will 20mhz actually have a noticeable difference for the applications we use and testing we do?
> 
> My only question before merging. I'm guessing the answer is no.

No
1) Most of the nodes are not meshed over meshpoint anyway
2) You would require a large bandwidth load be notice the slowness in the session

You may actually see a performance increase for lower-bandwidth applications especially in saturated areas.

makeworld-the-better-one:
Sounds good let's do it then, unless @benhylau has an objection.

darkdrgn2k:
My biggest worry is backwards compatibility.

makeworld-the-better-one:
It is prototype though. You shouldn't expect complete compatibility. We can mention that in won't be backwards compatible when we do a new release, so it's at least not a surprise, put it in the README or something.

darkdrgn2k:
You can easily change older builds to be compatible.
(remove HT40+ from `sudo iw dev $mesh_dev ibss join MESH_NAME 2412 HT40+`)
or newer builds to work with older ones
(Add HT40+ to `sudo iw dev $mesh_dev ibss join MESH_NAME 2412`)

makeworld-the-better-one:
Then as long as this gets documented, maybe in the new MODULES.md file, I feel good about merging.

dasanchez:
@darkdrgn2k just to clarify, the proposed change can be stated as
**Make both Access Point and Mesh Point operate on the 20MHz band by default**
?

darkdrgn2k:
@dasanchez 
Currently AP is 20 Mhz

Meshpoint and AdHoc is set to 40Mhz

Proposition is to move MeshPoint for 20 Mhz as well

This would also allow onboard wifi of raspberry pis to join a default ad-hoc network as well (if not set to ap)

PR #274, opened 2019-01-31
Added check for cjdns tunnel interface

### LINKED ISSUE #247: Status hangs when cjdns no started

darkdrgn2k:
fixes #247 

Added check for tun0 actually being available before reporting peers

### DIFF
--- a/scripts/status
+++ b/scripts/status
@@ -106,7 +106,7 @@ if [ "$(which yrd)" ]; then
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


darkdrgn2k:
> 
> 
> This looks fine, would it be better to check the status of the service though?

Possibly, unless the service is not running (manually started cjdns)

but welcome to update it if some one wants to

PR #276, opened 2019-01-31
SSB neighbour ping

darkdrgn2k:
Sends packets to neighbors regarding ssb existence


darkdrgn2k:
> 
> 
> Looks good, I'd need testing to know for sure. I assume you tested it though.

In an isolated environment.

Will test more in devel

PR #279, opened 2019-01-31
(Minor, before release) Script to configure github quickly

darkdrgn2k:
Don't know if this is a desired thing in contrib.

I found its a pain to have to configure my git hub identity on the pi every time i rebuild it so i can commit changes.

Script sets up your git environment so you can push changes from the pi.

👍  or 👎 

*note* CI showing unrelated error

### DIFF
--- a/contrib/github-setup/setid.sh
+++ b/contrib/github-setup/setid.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+if [ -z "$1" ] && [ -z "$2" ]; then
+   echo Syntax 
+   echo setid \[git-hub login\]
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
Idk how much I'll use it, but there's no harm, especially because it's in contrib.

darkdrgn2k:
> Idk how much I'll use 

if you want to git commit/push right from the pi you need to set these variables

makeworld-the-better-one:
Ok, sure. Let's merge.

benhylau:
Either merge or close this?

Issue #283, opened 2019-02-07
Only peer with direct mesh peers


makeworld-the-better-one:
Many services, such as IPFS, might benefit from only peering with mesh peers, instead of hundreds of indirect peers across the mesh. Those direct peers can peer further and further, but IPFS won't try and teach too far.

makeworld-the-better-one:
SSB (with the exception of pubs) and IPFS (possibly with the exception of a starter DHT) could do this, but it also might mess things up heavily. Something to experiment with.

makeworld-the-better-one:
With the rename of this issue, I would add that this isn't just about direct mesh peers, it's also about stuff like IPFS trying to connect to clearnet nodes when it doesn't have access, or DHT splitting, stuff like that.

PR #284, opened 2019-02-08
IPFS Improvements

makeworld-the-better-one:
This PR adds several experimental IPFS improvements. It needs testing.
They are:

- Relay hopping is enabled
    - > If EnableRelayHop is enabled, the node will act as an intermediate (Hop Relay) node in relay circuits for connected peers. What does all this mean? Circuit relaying provides peers with the means to indirectly connect *other* peers who cannot directly connect to each other, either because of NAT or because of protocol incompatibilities, such as browser-based (`js-ipfs`) peers connecting to desktop (`go-ipfs`) peers. This is a nice feature to have if you want to be able to relay information for browser-based Dapps.
    - It seems it will help the overall health of the network
    - *REMOVED*
- NAT port mapping is disabled as NATs do not come into play here
    - Perhaps this should be renabled, as sometimes at homes we will have these nodes under NAT?
    -  *REMOVED*
- Swarm connection management was setup
    - `LowWater`
        - Lowest # of connections that it attempts to maintain
        - At 600 by default, I changed it to 100
    - `HighWater`
        - Highest number of connections before they are discarded
        - At 900 by default, I changed it to 200
    - `GracePeriod`
        - How long connections last before they can be closed, even when HighWater has been reached. Prevents abrupt closure.
        - > In low power situations, you might want to kick the `GracePeriod` up to 1 minute, but drastically reduce the `LowWater` and `HighWater` values.
        - At 30s by default, I changed it to 60s
    - **These values need testing to be set properly, and I'd like your input**
- `ipfs-swarm.sh` was renamed to `ipfs-setup.sh`
    - The rename shows up like I've deleted one file and entirely created another, unfortunately
- Swarm filtering was added
    - This is a CIDR blacklist, meaning IPFS will not try to connect to nodes with addresses within the blacklisted subnet, even if they show up in the DHT and stuff.
    - This is built into IPFS
    - `ipfs-setup.sh` checks if CJDNS, Yggdrasil, IPv4 Internet and IPv6 Internet are available, and blacklists them if they are not
    - This will hopefully make IPFS more efficient, and reduce the issues presented in #283 
    - It's [now](https://github.com/tomeshnet/prototype-cjdns-pi/pull/284#issuecomment-461997424) a loop
- [QUIC](https://en.wikipedia.org/wiki/QUIC) aka HTTP/3
    - QUIC uses UDP and aims to replace TCP
    - Less latency, attempts to avoid congestion
    - IPFS has [experimental support](https://github.com/ipfs/go-ipfs/blob/master/docs/experimental-features.md#quic) for connecting to peers with it
    - This PR enables support for it
    - Adds a `quic_enabled` bool to nodeinfo.json
    - `ipfs-setup.sh` will try and use QUIC to peer with mesh peers, if they say they have enabled it in their nodeinfo
    - If you guys agree with supporting QUIC, then we could connect to the default CJDNS and Ygg peers over QUIC as well


scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,102 @@
...
+    ipfs swarm filters add '/ip6/fc00::/ipcidr/8'
+    echo "Blocked CJDNS network"
+fi
+# Yggdrasil
+if ! [ "$(systemctl status yggdrasil.service | grep 'Active: ' | awk '{ print $2 }')" = 'active' ]; then
+    ipfs swarm filters add '/ip6/0200::/ipcidr/7'
+    echo "Blocked Yggdrasil network"
+fi
+# Clearnet or regular Internet IPv4 access
+if ! [ "$(ping -c 3 1.1.1.1 &> /dev/null)" ]; then
```
darkdrgn2k: Not sure this works how you think.

Also in past experience this can return false data.

Also what if your tunnled over hyperbola (IP TUNNEL)? Do we care?

```
[root@localhost 1]# if ! [ "$(ping -c 3 99.99.99.99 &> /dev/null)" ]; then echo Good; fi
Good
[root@localhost 1]# if ! [ "$(ping -c 3 1.1.1.1 &> /dev/null)" ]; then echo Good; fi
Good
```

darkdrgn2k:
Allot of these changes assume that we will be running a MESH ONLY network which is the exact opposite.

Additionally the ipfs-setup script does not take into account the fact that i may plug my computer into a my wired home network after the fact.

I think i'd be more comfortable with some sort of toggle to enable/disable internet peers instead of testing for internet.

scripts/ipfs/install
```diff
@@ -29,9 +29,25 @@ ipfs init || true
...
 ipfs config Pubsub.Router gossipsub
 
+# Enable relay hop to help peers with connecting
+ipfs config --bool Swarm.EnableRelayHop true
+
 # Enable Filestore for --nocopy capability
 ipfs config --bool Experimental.FilestoreEnabled true
 
+# Disable NAT port mapping used for NAT traversal
+ipfs config --bool DisableNatPortMap true
```
darkdrgn2k: What about pis connected to the home internet, your killing their ability to try to gain more peer by nat port.

scripts/ipfs/install
```diff
@@ -29,9 +29,25 @@ ipfs init || true
 # Enable gossipsub routing
 ipfs config Pubsub.Router gossipsub
 
+# Enable relay hop to help peers with connecting
+ipfs config --bool Swarm.EnableRelayHop true
```
darkdrgn2k: This would increase bandwidth usage on internet connected peers.

Also how does this work when you block internet peers.  Will that mean that a non-internet enabled peer will not want to relay through an internet relayed one?

scripts/ipfs/ipfs-setup.sh
makeworld-the-better-one:
```
❯ ping -w 2 99.99.99.99 &>/dev/null; if [ $? -eq 0 ]; then echo hello; fi

❯ ping -w 2 1.1.1.1 &>/dev/null; if [ $? -eq 0 ]; then echo hello; fi
hello
```
I will add this fix.

scripts/ipfs/install
makeworld-the-better-one: Yeah, I mentioned that in my description but didn't change it. I will change it back.

scripts/ipfs/install
makeworld-the-better-one: > This would increase bandwidth usage on internet connected peers.

Possibly, but only if we manually set up the relay to be used. The current release does not automatically use relays. With this, we have the option too, and it will improve the overall health of the network.

> Also how does this work when you block internet peers. Will that mean that a non-internet enabled peer will not want to relay through an internet relayed one?

As was said in chat by me:
> Whether it will allow the connection to a blacklisted peer over a relay is a good question though, something to be tested. Just keep in mind IPFS won't try to do that on its own just yet

Relaying through an Internet peer would actually be a good thing for nodes to do, but we will have to wait until IPFS does it automatically, possibly in the next release. Once that happens, we can likely remove all the filtering entirely.

scripts/ipfs/ipfs-setup.sh
makeworld-the-better-one: After shellcheck, I am going with:
```
if ! ping -w 2 1.1.1.1 &>/dev/null; then
```

makeworld-the-better-one:
In regards to `LowWater` and `HighWater` values, the IPFS lowpower profile sets them to 20 and 40 respectively. I hope that provides some reference, but to be honest I have no idea what it means for the Pis. Formal testing of resource use and performance with different values is probably required.

makeworld-the-better-one:
I made the script a loop, which is a big, potentially stupid change. To see how this changes what I've done so far, look at the diff [here](https://github.com/tomeshnet/prototype-cjdns-pi/pull/284/commits/439a60e05dccc21efa54d8ece9c0ef3e07fea40f).

makeworld-the-better-one:
`develop` will need to be merged back into this branch after the release.

makeworld-the-better-one:
This PR should update to the [0.4.19 release](https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#0419-2018-03-01) now instead, and change some of the settings accordingly.

darkdrgn2k:
@makeworld-the-better-one i would suggest looking at the latest release and doing a version bump

makeworld-the-better-one:
@darkdrgn2k The latest IPFS release? Yes, that's the comment I made just above. I will also remerge from develop.

makeworld-the-better-one:
I am going to move on to the testing stage for this PR, but before/as I do, does anyone (@benhylau and @darkdrgn2k) have issues with the script being a loop that blocks access to certain groups of peers as the environment changes? I'm not requesting a review yet.

I have confirmed that even with the latest IPFS release, you cannot access hidden networks (Yggdrasil on a CJDNS-only node, for example) by *automatically* using a relay. Manually connecting through a relay to access these networks could be possible, but should be done in a different PR.

darkdrgn2k:
> being a loop that blocks access to certain groups of peers as the environment changes?

I can tell you from experience that something like "is there internet" by pinging 8.8.8.8 cannot guarantee an internet connection.  Just the other day in the chat Ikata incorrectly though his internet was not working when 8.8.8.8 did not respond to pings.

Additional your code will block an lan peering if there is no internet.  You also cant assume that private ip addresses will be used.

Finally  there has not been any established baseline of blacklisting a peer vs simply letting it fail the connection.

If you wish to peruse this method anyway i would suggest the following
- Split it of into a separate optional service (probably separate module too)
- Make sure you add a delay in the loop, you dont want it being seen as a d-dos
- Be careful with overusing `/opt/cjdns/publictoip6`  This is a process intensive function
- I would NOT put the peering support into a loop, especially in the means you have done it

Remember also that we ill need to separate the yggdrasil and cjdns stuff away from the main ipfs module to make it more "generic" for DEBifying.

darkdrgn2k:
Comment from IPFS channel

>presumably the attempts to connect (incl bootstrap) via internet addresses will eventually time out. i guess blacklist will save on this delay, but only if all parallel attempts are taken up by the internet addressses, starving the cjdns/other addresses. 

I think the key here was `but only if all parallel attempts are taken up by the internet addresses`

This means that there are so many internet peers trying to be contacted that there i no room for cjdns ones. I'm pretty sure this number is high so makes me question the effectiveness of this hack in the grand scheme of things.

scripts/ipfs/install
```diff
@@ -38,6 +38,16 @@ ipfs config Pubsub.Router gossipsub
 # Enable Filestore for --nocopy capability
 ipfs config --bool Experimental.FilestoreEnabled true
 
+# Setup connection management - Reduce connections to stress the Pi less
+# XXX: These values need to be tweaked and tested
```
benhylau: @makeworld-the-better-one How will you measure whether these values provide an improvement or otherwise?

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+
+# This script connects to local mesh peers,
+# and it sets up connection filters based on what networks this node can access.
+# It runs continually, to change IPFS settings as the environment around the node changes.
+
+
+# Wait for IPFS to initalize
+attempts=15
+until [[ $(curl http://localhost:5001/api/v0/id -s 2>/dev/null) || ${attempts} -eq 0 ]]; do
+    sleep 1
```
benhylau: I don't think we need to test so often and I don't think 15 s is enough in some cases?

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+        if [[ ! ${id} == "null" ]] && [[ ! "${id}" == "" ]]; then
+            # Connect to neighbouring IPFS nodes
+            # Check for QUIC connections first
+            if [ "$(echo "${res}" | jq -r -M '.services.IPFS.quic_enabled')" == 'true' ]; then
+                # ID is not needed for QUIC connections
+                ipfs swarm connect "/ip6/${addr}/udp/4001/quic"
+                echo "Connecting to ${addr} with QUIC"
+            else
+                ipfs swarm connect "/ip6/${addr}/tcp/4001/ipfs/${id}"
+                echo "Connecting to ${addr} regularly"
```
benhylau:
```suggestion
                echo "Connecting to ${addr} over TCP"
```

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+    res=$(curl http://["${addr}"]/nodeinfo.json -s)
+    if [ ! -x "${res}" ]; then
+        id=$(echo "${res}" | jq -r -M '.services.ipfs.ID')
+        # Value is found
+        if [[ ! ${id} == "null" ]] && [[ ! "${id}" == "" ]]; then
+            # Connect to neighbouring IPFS nodes
+            # Check for QUIC connections first
+            if [ "$(echo "${res}" | jq -r -M '.services.IPFS.quic_enabled')" == 'true' ]; then
+                # ID is not needed for QUIC connections
+                ipfs swarm connect "/ip6/${addr}/udp/4001/quic"
```
benhylau: I think the log should come before the command

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+        done <<< "$(sudo nodejs /opt/cjdns/tools/peerStats 2>/dev/null | awk '{ if ($3 == "ESTABLISHED") print $2 }' | awk -F. '{ print $6".k" }' | xargs)"
+
+        # Remove a filter if there was one
+        ipfs swarm filters rm '/ip6/fc00::/ipcidr/8'
+        echo "Enabled access to CJDNS network"
+    else
+        ipfs swarm filters add '/ip6/fc00::/ipcidr/8'
+        echo "Blocked CJDNS network"
+    fi
+
```
benhylau:
```suggestion

    # Check if Yggdrasil is running
```

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+    if [ "$(systemctl status yggdrasil.service | grep 'Active: ' | awk '{ print $2 }')" = 'active' ]; then
+        while read -r ygg_peer; do
+            addPeer "${ygg_peer}"
+        done <<< "$(sudo yggdrasilctl getPeers | grep -v "(self)" | awk '{print $1}' | grep -v bytes_recvd | xargs)"
+
```
benhylau:
```suggestion

        # Remove a filter if there was one
```

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+        echo "Enabled access to Yggdrasil network"
+    else
+        ipfs swarm filters add '/ip6/200::/ipcidr/7'
+        echo "Blocked Yggdrasil network"
+    fi
+
+    # Clearnet or regular Internet IPv4 access
+    if ! ping -w 2 1.1.1.1 &>/dev/null; then
+        # Block all IPv4 - it's not used anywhere else
+        ipfs swarm filters add '/ip4/0.0.0.0/ipcidr/0'
```
benhylau: What if I am on a off-Internet LAN local meshnet plugged into my local switch? Now my IPFS won't swarm?

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+
+        ipfs swarm filters rm '/ip6/200::/ipcidr/7'
+        echo "Enabled access to Yggdrasil network"
+    else
+        ipfs swarm filters add '/ip6/200::/ipcidr/7'
+        echo "Blocked Yggdrasil network"
+    fi
+
+    # Clearnet or regular Internet IPv4 access
+    if ! ping -w 2 1.1.1.1 &>/dev/null; then
```
benhylau: This type of Internet check is usually problematic. What is the motivation behind adding this filter at all?

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+        # Block all IPv4 - it's not used anywhere else
+        ipfs swarm filters add '/ip4/0.0.0.0/ipcidr/0'
+        echo "Blocked all of IPv4"
+    else
+        ipfs swarm filters rm '/ip4/0.0.0.0/ipcidr/0'
+        echo "Enabled access to all of IPv4"
+    fi
+
+    # Clearnet IPv6 access
```
benhylau: Same as IPv4. Recommend removing both.

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+    if ! ping -w 2 2606:4700:4700::1111 &>/dev/null; then  # IPv6 version of 1.1.1.1 above
+        # Block global unicast for IPv6
+        ipfs swarm filters add '/ip6/2000::/ipcidr/3'
+        echo "Blocked IPv6 internet"
+    else
+        ipfs swarm filters rm '/ip6/2000::/ipcidr/3'
+        echo "Enabled access to IPv6 internet"
+    fi
```
benhylau: I would at least so a 60 s sleep on this loop. Probably more.

scripts/ipfs/ipfs-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+            else
+                ipfs swarm connect "/ip6/${addr}/tcp/4001/ipfs/${id}"
+                echo "Connecting to ${addr} regularly"
+            fi
+        fi
+    fi
+}
+
+# Continually mesh with peers and add or remove filter rules as networks go up and down
+while true; do
```
darkdrgn2k: I dont know how `postexec` will work for long running processes. Upon service restart I dont think the PID will be killed but it will spawn again.

This will created multiple scripts running in parallel.

Better off running in a service

makeworld-the-better-one:
Thanks for the comments on the PR! 

I will edit it and remove the loop, and instead just leave the tweaking of some settings and doing a version update.

scripts/ipfs/install
makeworld-the-better-one: CPU and RAM usage can be looked, as well as file retrieval times and IPNS resolve. It will require a formal test, but I know that these values will be better than the ones we have now, especially since this is on such limited hardware.

makeworld-the-better-one:
I've now updated the PR, it is much smaller now. Here are the changes as compared to `develop`

- Version update to 0.4.19
- Set config so that the daemon connects with fewer peers, to save on resources
- Added connecting to peers over QUIC, if they have it enabled

PR #288, opened 2019-02-12
Status - added sudo to support debin

darkdrgn2k:
i thing Debian doesnt have /sbin in the path of normal users, sudo is a work around to this

### DIFF
--- a/scripts/status
+++ b/scripts/status
@@ -24,7 +24,7 @@ fi
 if [ "$(which mesh-point)" ]; then
     ints=$(sudo iw dev | grep Interface | awk '{print $2}')
     while read -r line; do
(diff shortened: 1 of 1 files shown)


scripts/status
```diff
@@ -22,9 +22,9 @@ else
 fi
 
 if [ "$(which mesh-point)" ]; then
-    ints=$(iw dev | grep Interface | awk '{print $2}')
+    ints=$(sudo iw dev | grep Interface | awk '{print $2}')
     while read -r line; do
-        mp=$(iw dev $line info | grep "type mesh point")
+        mp=$(sudo w dev $line info | grep "type mesh point")
```
makeworld-the-better-one: Should read as `sudo iw`.

PR #289, opened 2019-02-12
Move functions to seperate file

darkdrgn2k:


makeworld-the-better-one:
Why do this? And what about other functions, like the one in `scripts/ipfs/ipfs-swarm.sh`?

darkdrgn2k:
> Why do this?
Because makes the install file allot more readable without these lines
Also makes these functions reusable in other modules without re-writing

> And what about other functions, like the one in `scripts/ipfs/ipfs-swarm.sh`?
Those are specific to ipfs-swarm.sh and dont really make the file bulky or unreadable

Issue #180, opened 2018-08-15
Avoid  in-place sed replacements, use temp files instead


darkdrgn2k:
REF: https://github.com/tomeshnet/prototype-cjdns-pi/pull/178#pullrequestreview-145879233

There are many places where `sed` is used to replace text in files.  The templated file with place holder is placed in its final location and then sed applies the changes.  To prevent the script failing and leaving a file with place holders, files should be modified in a temporary location before being moved to their final resting place.

PR #294, opened 2019-02-19
Added set-mesh-repo script to contrib

ASoTNetworks:
Added contrib script for setting up mesh mirrors for APT.

### DIFF
--- a/contrib/set-mesh-repo/armbian-armbian.list
+++ b/contrib/set-mesh-repo/armbian-armbian.list
@@ -1,0 +1,1 @@
+deb http://__PREFIX__.mirror.tomesh.net/armbian/apt/ __CODENAME__ main __CODENAME__-utils __CODENAME__-desktop
--- a/contrib/set-mesh-repo/set-repo.sh
+++ b/contrib/set-mesh-repo/set-repo.sh
@@ -83,7 +83,6 @@ then
 		cp armbian-armbian.list /etc/apt/sources.list.d/armbian.list
 		sed -i "s/__CODENAME__/$codename/g" /etc/apt/sources.list.d/armbian.list
 		sed -i "s/__PREFIX__/$prefix/g" /etc/apt/sources.list.d/armbian.list
-		
 	fi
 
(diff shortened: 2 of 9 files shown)


darkdrgn2k:
Maybe sed against a temp file and then move it over so that if it breaks during sed we dont have a broken config sitting in /etc

ref: https://github.com/tomeshnet/prototype-cjdns-pi/issues/180

contrib/set-mesh-repo/set-repo.sh
```diff
@@ -0,0 +1,94 @@
...
+	echo "Changing to Debian $1 repo."
+	cp debian-sources.list /etc/apt/sources.list
+	
+	# check if there is /etc/apt/sources.list.d/armbian.list if so replace it
+	if [ -f /etc/apt/sources.list.d/armbian.list ]
+	then
+		cp armbian-armbian.list /etc/apt/sources.list.d/armbian.list
+		sed -i "s/__CODENAME__/$codename/g" /etc/apt/sources.list.d/armbian.list
+		sed -i "s/__PREFIX__/$prefix/g" /etc/apt/sources.list.d/armbian.list
+
```
benhylau:
```suggestion
```

benhylau:
Note this would only work if internet peers are added so we can reach main nets.

benhylau:
Can we add this mirror URL to this list before merging? https://github.com/tomeshnet/documents/pull/113/files

contrib/set-mesh-repo/armbian-armbian.list
```diff
@@ -0,0 +1 @@
+deb http://__PREFIX__.mirror.tomesh.net/armbian/apt/ __CODENAME__ main __CODENAME__-utils __CODENAME__-desktop
```
benhylau: I think we should use `__PREFIX__.mirror.debian.tomesh.net` or something similar, otherwise when we change this eventually to reclaim the `mirror.tomesh.net` we'd end break existing installs.

PR #304, opened 2019-02-21
add kill wpa-supplicant to adhoc script

darkdrgn2k:
wpa_supplicant sometimes interferes with adhoc preventing the interface from coming up

so kill it


makeworld-the-better-one:
Did you mean to merge this into develop instead?

makeworld-the-better-one:
Is there a way in can be restarted after ad-hoc is done being used?

PR #305, opened 2019-02-22
yggdrasil IP Tunnel

darkdrgn2k:
Parallels CJDN IP Tunnels with the following exception

- yggdrasil has no way to change config during run time. yggdrasil-pre-setup script was added to re-write the config file by converting it to json and adding the config entries needed to setup the tunnels needed before yggdrasil starts

- yggdrasil does not configure the network automatically.  this means that ip addresses need to be decided on before hand.  This means the yggdrasil.iptunnel.* contain two files. PUBKEY and an IP ADDRESS. These two are separated by a space. server does  not automatically index it because the remote side ip address must match.

- Because manual network configuration is needed default route is created manually to.  This means only 1 default route per client.

- Default route is a bit of a hack to preserve the original route. Instead of overwriting the existing default route causing connection to die if yggdrasill ever goes down, it creates 2 "more specific" routes that override the 1 more generic route but still over the whole internet. (openvpn uses this hack)


docs/MODULES.md
```diff
@@ -25,6 +25,8 @@ A short summary of each module is directly below. Documentation for specific abi
...
 | `WITH_YGGDRASIL`                | None                                           | Set to `true` if you want to install [Yggdrasil](https://yggdrasil-network.github.io/), an alternate and possibly more efficient mesh routing software than CJDNS. |
+| `WITH_YGGDRASIL_IPTUNNEL`       | None                                           | Set to `true` if you want to use the yggdrasil iptunnel feature to set up an Internet gateway for your node. To configure as a server (exit Internet traffic for other nodes), create **/etc/yggdrasil.iptunnel.server** containing a newline-separated list of cjdns public keys of allowed clients and an ipaddress for that client. To configure as a client (use an exit server to access the Internet), create **/etc/yggdrasil.iptunnel.client** containing a newline-separated list of cjdns public keys of the gateway servers followed by the IP address set on the server. You can only configure as one or the other, not both and you can only have one entry on the client. |
```
makeworld-the-better-one: This is a pretty long line. Maybe give a one sentence explanation of what it does, not how to do it, and then link to the doc section below?

scripts/yggdrasil-iptunnel/yggdrasil-pre-setup
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env bash
+
+# Convert existing yggdrasil config to json from hjson
```
makeworld-the-better-one: Why is this only being done in `yggdrasil-iptunnel`? Why is it an issue in this case, and not with regular Yggdrasil?

scripts/yggdrasil-iptunnel/yggdrasil-pre-setup
```diff
@@ -0,0 +1,47 @@
...
+   
+# Detect if there is a # as a first character of the file.
+# If it is then its proably a hjson and needs to be converted into json
+if [ ! -z "$(tr -d ' \t\r\f' < /etc/yggdrasil.conf   | grep -i "^[#]")"  ]; then
+    sudo mv /etc/yggdrasil.conf /etc/yggdrasil.conf.orig
+    sudo yggdrasil -useconffile /etc/yggdrasil.conf.orig -normaliseconf -json | sudo tee /etc/yggdrasil.conf > /dev/null
+fi
+
+if [ -e /etc/yggdrasil.iptunnel.server ]; then
+
```
benhylau: Why these extra newlines?
```suggestion
```

scripts/yggdrasil-iptunnel/yggdrasil-pre-setup
```diff
@@ -0,0 +1,47 @@
...
+            IPv4Destinations="${IPv4Destinations} \"${IP_ADDR}/32\": \"${PUBLIC_KEY}\","
+        fi
+    done < /etc/yggdrasil.iptunnel.server
+
+    # Trim last ,
+    IPv4Destinations="${IPv4Destinations%?}"
+    IPv4Sources="0.0.0.0/0"
+
+elif [ -e /etc/yggdrasil.iptunnel.client ]; then
+
```
benhylau:
```suggestion
```

scripts/yggdrasil-iptunnel/yggdrasil-setup
```diff
@@ -0,0 +1,43 @@
...
+    
+    # If no NAT (masquarading) being done, start doing it
+    if [ -z "$(sudo iptables -L POSTROUTING -v -n -t nat  | grep MASQUERADE)" ]; then
+        sudo iptables -t nat -A POSTROUTING -o eth0 -j MASQUERADE
+    fi
+
+# Configure as yggdrasil iptunnel client if client file is present (this is simply a newline-separated list
+# of cjdns public keys in /etc/yggdrasil.iptunnel.client, each key indicating an iptunnel exit server)
+elif [ -e /etc/yggdrasil.iptunnel.client ]; then
+
```
benhylau:
```suggestion
```

scripts/yggdrasil-iptunnel/uninstall
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Uninstall scripts that configure cjdns iptunnel
```
benhylau:
```suggestion
# Uninstall scripts that configure yggdrasil iptunnel
```

scripts/yggdrasil-iptunnel/uninstall
```diff
@@ -0,0 +1,13 @@
...
+sudo rm /usr/local/sbin/yggdrasil-setup
+sudo rm /usr/local/sbin/yggdrasil-pre-setup
+if [ -f /lib/systemd/system/cjdns.service ]; then
```
benhylau: I am guessing this should be yggdrasil?

```suggestion
if [ -f /lib/systemd/system/yggdrasil.service ]; then
```

scripts/yggdrasil-iptunnel/uninstall
```diff
@@ -0,0 +1,13 @@
...
+    sudo sed -i /ExecStartPost/d /lib/systemd/system/yggdrasil.service
+fi
+if [ -f /etc/systemd/system/yggdrasil-pre-setup ]; then
```
benhylau: Should this be the service?

scripts/yggdrasil-iptunnel/install
```diff
@@ -0,0 +1,22 @@
...
+
+sudo apt-get -y install jq
+
+sudo cp "$BASE_DIR/yggdrasil-setup" "/usr/local/sbin/yggdrasil-setup"
+sudo cp "$BASE_DIR/yggdrasil-pre-setup" "/usr/local/sbin/yggdrasil-pre-setup"
+
+sudo chmod a+x /usr/local/sbin/yggdrasil-pre-setup
+sudo chmod a+x /usr/local/sbin/yggdrasil-setup
+
+# Update service to start script on cjdns start
```
benhylau: cjdns or yggdrasil? I suggest searching for all instances of cjdns in this diff to make sure they are not misused for places where it should say yggdrasil.

docs/MODULES.md
benhylau: This line contains "cjdns" in place of yggdrasil. Please check all instances.

darkdrgn2k: Checked

scripts/yggdrasil-iptunnel/yggdrasil-pre-setup
darkdrgn2k: because you cannot cleanly edit hjson.  (IE jq does not work)

PR #311, opened 2019-02-26
NodeExporter ygg rx/tx

darkdrgn2k:
Provides Yggdrasil peer information and RX/TX data to node exporter if installed

Need to add condition to check if yggdrasilctl errored out

check for valid json or  string `Fatal error:`


darkdrgn2k:
Two check to make sure parsable  content it returned

- [ ] Add json check for cexec
- [ ] Add json check for yggdrasilctl

PR #312, opened 2019-02-26
PeerDNS

makeworld-the-better-one:
Adds a contrib module for [PeerDNS](https://github.com/p2pstuff/PeerDNS/), an experimental DNS system that is decentralized, permissionless, designed for mesh, and uses trust values to reconcile double entries. It has a Web UI, and solves the DNS stuff brought up in our HackMD. CJDNS support built-in by the dev!

Should this be a regular module? The software is still alpha, and doesn't have formal releases yet, but does work. If it becomes a regular module, I'll add docs for it formally.


makeworld-the-better-one:
I will be changing the systemd setup according to [this suggestion](https://matrix.to/#/!ejHayYuUWojLYzdTEP:m.trnsz.com/$1551167999151698ESHnM:matrix.org?via=m.trnsz.com&via=matrix.org&via=tomesh.net) from the dev.

makeworld-the-better-one:
I will also be making this a full fledged module with docs, an nginx reverse-proxy, and more - for after the release.

makeworld-the-better-one:
This module PR will be reopened once most of RC v0.5 is complete, and debs are working.

Issue #165, opened 2018-07-26
Identify why the -O2 is required


darkdrgn2k:
More and more boards need -O2 to be used to compile CJDNS

Could be a GCC7 thing.

PR #340, opened 2019-03-10
Support for x86 (32bit and 64bit)

### LINKED ISSUE #204: Look into grafana version bump
Grafana now at version 5.x

Bump and comfirm
- [x] Installs fine
- [x] Compatible with Prometheus
- [x] Dashboard works

Also check Prometheus and node-exporter versions

note: ARM64 version only exists in version v5+

### LINKED ISSUE #212: prometheus-node-exporter version bump
new version of  prometheus-node-exporter need ` --collector.textfile.directory` instead of  "-collector.textfile.directory"

### LINKED ISSUE #328: Add i386 Support
Add i386 support for all modules

This will make testing easier and open the project to virtual machines and such.

### LINKED ISSUE #329: Rewrite board detection
Related to #292

Simplify board detection and roll it into reusable functions

darkdrgn2k:
New board detection routine Closes #329
Added support for x86 and arm64 Closes  #328
Node Exporter Closes #212 , Grafana Closes #204 and Prometheus version bump

- [x] FULL install on Debian 9 VM
- [x] It keeps wanting to access ipv6 addresses when ipv6 is not around, maybe demote ipv6 by default? (updated in ygg-tunnel-ipv6 branch)
- [x] Test new detection method on some other boards (probably will have to happen after merge)
- [x] Add support for i686 (same as i386)


scripts/grafana/install
```diff
@@ -2,21 +2,37 @@
...
+  ;;
+  i386 | i586 | i686 )
+   echo "Cannot install grafana on i386"
+   exit 0
+  ;;
+  armv7l)
+    URL="https://dl.grafana.com/oss/release/grafana_${GRAFANA_VERSION}_armel.deb";
+  ;;
+  armv6l)
+    # Use old grafana - no new one available
```
makeworld-the-better-one: Maybe have a separate `GRAFANA_ARMHF_VERSION` variable to change later?

scripts/grafana/install
```diff
@@ -2,21 +2,37 @@
...
+    URL="https://bintray.com/fg2it/deb-rpi-1b/download_file?file_path=main%2Fg%2Fgrafana_5.1.4_armhf.deb"
+  ;;
+  aarch64)
+    URL="https://dl.grafana.com/oss/release/grafana_${GRAFANA_VERSION}_arm64.deb";
+  ;;
+  *)
+    echo "Unknown Arch"
+    exit 1
```
makeworld-the-better-one: You use `exit 0` up above for when you can't download because of i386. Does the code make any difference?

scripts/functions.sh
```diff
@@ -127,3 +127,65 @@ function askSelection {
         dialogREPLY="$REPLY"
     fi
 }
+
+function detectBoard {
+
+    BOARD_MODEL="Unknown"
+    BOARD_NAME="Unknown"
+    BOARD_OS="Unknown"
```
makeworld-the-better-one: Are these variables exported?

scripts/install2
```diff
@@ -3,134 +3,99 @@
...
-# Use tr to remove null byte generating warning
-BOARD_MODEL=$(tr -d '\0' < /sys/firmware/devicetree/base/model)
-BOARD_HARDWARE=$(grep Hardware /proc/cpuinfo | awk '{print $3}' | head -n 1)
+detectBoard;
 
 
+# Set some defaults
+if [[ "$ARCH" == "armv7l" ]]; then
+    #TODO# -O2 workaround. Needs to be resolved.
+    CJDNS_BUILD_CMD="sudo Seccomp_NO=1 CFLAGS=\"-O2 -s -static -Wall -march=armv7-a -mtune=cortex-a7 -mfpu=neon-vfpv4 -mfloat-abi=hard -fomit-frame-pointer -marm\" ./do"
```
makeworld-the-better-one: Maybe single quotes would make this cleaner?

scripts/yggdrasil/install
```diff
@@ -5,18 +5,38 @@ set -e
...
 BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
+ARCH="$(uname -m)" 
+case "$ARCH" in
+  x86_64)
+    ARCH_VERSION="amd64"
+  ;;
+  i386 | i586 | i686 )
+   ARCH_VERSION="i386"
+   exit 0
```
makeworld-the-better-one: Yggdrasil does have i386 support, why does it exit?

scripts/yggdrasil/install
```diff
@@ -5,18 +5,38 @@ set -e
...
+  ;;
+  armv7l)
+   ARCH_VERSION="armhf"
+  ;;
+  armv6l)
+    # Use old grafana - no new one available
```
makeworld-the-better-one: Copied over comment.

scripts/functions.sh
darkdrgn2k: No.

Instead of being exported  somewhere in the script then used by the modules, the function is part of a shared file and the function  can be called by any module when needed to set it in their environment. 

This allows each module to be run without having to run the whole script to set them and be independent.

scripts/grafana/install
darkdrgn2k: I have not been able to find any other Grafana for armv6l and i dont think it will be updated.
If some one can find it please let me know

Otherwise the options are

- Not have Grafana available
- Keep it static to an old version as it is now :( (at least for now still compatible)
- Build it ourselves (any takes on building it from source :))

We can Create this as an issue if there is interest in it

PS: armv6l = Raspbery Pi 1 and Zero

scripts/grafana/install
darkdrgn2k: Difference is as folows:

0 will exit the script but continue executing install  (we KNOW i386 is not available so  just continue)
1 will stop execution the install script (Something went wrong if we dont know what ARCH it is!)

scripts/install2
darkdrgn2k: Maybe. But its a copy /paste from Prototype version1.

The whole cjdns build needs to be assessed and cleaned up Noted in #165

darkdrgn2k:
> Looks like great work to support more boards, thanks a lot! I left some small comments. The only large thing I think is left is updating docs to explain the support that is not offered, and the variety of boards that fall under it. I think that should be done in this PR.

Grafana note added in this repo
Prometheus in Prometheus Doc branch to avoid conflicts in mearning

scripts/functions.sh
makeworld-the-better-one: Ah I see, so the variables are set every time.

scripts/grafana/install
makeworld-the-better-one: Building a binary ourselves would be the best for compatibility, but there are trust issues, since they won't be confirmed by the devs. Maybe you can contact the devs about this (make a new issue for it in the repo too), and for now we'll leave the old version? I would also some sort of warning with `echo`.

PR #341, opened 2019-03-11
IPv4 over IPv6 Resolution

darkdrgn2k:
Sometimes the system will think that oyu have ipv6 when yggdrasill / cjdns are active. This casues some sites that have both an A and AAAA record not to load  (because it tries the AAAA record)

This script changes gia.conf to lower ipv6 address preferences below ipv4 making ipv4 preferred

gai stands for getaddrinfo, the standard system call for resolving host names

### DIFF
--- a/contrib/ipv4-use-first/ipv4-first
+++ b/contrib/ipv4-use-first/ipv4-first
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+
+# gai stands for getaddrinfo, the standard system call for resolving host names
+# Change will use ipv4 over ipv6 when possible
+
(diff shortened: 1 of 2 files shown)


darkdrgn2k:
Ref #340 

Debian 9 seems to be affected during install

makeworld-the-better-one:
Why was this closed as well?

makeworld-the-better-one:
@darkdrgn2k This should be reopened and set to merge to the `develop` branch as well.

darkdrgn2k:
Marking as different class instead in in #309

PR #353, opened 2019-03-15
Document corrections

darkdrgn2k:
Corrected document errors


darkdrgn2k:
This issue affects master as well

- Change without releasing
- release v4.2
- do nothing

makeworld-the-better-one:
Hmm, I thought I fixed this when I made MODULES.md. Apparently not, or maybe it got reverted. Thanks for fixing it either way.

As for how to merge, I would change it in master without releasing. Broken links are bad, but don't merit a full blown release either.

darkdrgn2k:
Synced from master instead

PR #356, opened 2019-03-17
Added to MODULES.md

makeworld-the-better-one:
I've added a Table of Contents, SSB peering info, and links to the public peers repos for both CJDNS and Yggdrasil.

The TOC is generated with [this](https://marketplace.visualstudio.com/items?itemName=yzhang.markdown-all-in-one) VSCode add on, I would recommend it for markdown in general.

### DIFF
--- a/docs/MODULES.md
+++ b/docs/MODULES.md
@@ -2,6 +2,31 @@
 
 A short summary of each module is directly below. Documentation for specific abilities of modules, or reference commands are further below.
 
+## Table of Contents
+- [Modules Documentation](#modules-documentation)
+  - [Table of Contents](#table-of-contents)
+  - [Command-line flags](#command-line-flags)
+  - [CJDNS](#cjdns)
+  - [Yggdrasil](#yggdrasil)
+    - [Yggdrasil subnetting](#yggdrasil-subnetting)
+    - [Yggdrasil IPTunnel](#yggdrasil-iptunnel)
+      - [Server](#server)
+      - [Client](#client)
+  - [IPFS](#ipfs)
+  - [Firewall](#firewall)
+    - [Applying changes](#applying-changes)
+    - [IPv4](#ipv4)
+      - [Change open ports](#change-open-ports)
+    - [IPv6](#ipv6)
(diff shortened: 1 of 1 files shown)


docs/MODULES.md
```diff
@@ -186,10 +211,12 @@ If you use this rule, there is no point in having any other Yggdrasil client rul
...
 You can specify a protocol, but that would limit the ports that are open.
 
-## Adding deprecated.systems peer info into CJDNS and Yggdrasil (Optional)
+## CJDNS and Yggdrasil public peering (Optional)
 
 #### CJDNS
 
+Other peers can be found in the [this](https://github.com/hyperboria/peers) repo of peers. Try to connect to only a few peers, and ones that are close to where you live. One peer is already shown below for you.
+
 Go to [Deprecated Systems](https://deprecated.systems/) website. You will see the following information:
```
darkdrgn2k: I think the docs should be more generic and not be linked to tomesh resources. 

Specifically listing deprecated.systems here may be inappropriate. I dont think we want these resources flooded with random people by advertising it so publicly.

darkdrgn2k: @benhylau  thoughts?

makeworld-the-better-one: It was like this before of course, but there's no harm in discussing it now. I tend to agree, it's easily abused by putting it here. I would just link to the public sites as I have done, and whether @benhylau wants to put the deprecated.systems stuff in the public repo is up to him.

makeworld-the-better-one:
I have removed the deprecated.systems peering info in the CJDNS, Yggdrasil and SSB sections. I would suggest that the dep.sys pub is added to public peering though in SSB (and CJDNS?), as it is the only public pub that has mesh, that I know of at least.

darkdrgn2k:
@makeworld-the-better-one 
https://github.com/tomeshnet/prototype-cjdns-pi/pull/356/files#diff-86c4f95673d4319d89af160b557e42b1R248

One last place it exists other then that it looks good

darkdrgn2k:
Looks good lets merge!

thanks @makeworld-the-better-one

PR #378, opened 2019-04-13
Migrate confset to a file

darkdrgn2k:
migrated confset to config.inc


makeworld-the-better-one:
Where is the config file this references?

darkdrgn2k:
> Where is the config file this references?

Forgot to `git add` it.  Commit added to fix this  thanks

docs/MODULES.md
```diff
@@ -157,63 +157,73 @@ To use this module you must have it installed. You can check to see if the file
...
+Section `[general]`
 
-**IPv6nat**
-Disables masquerading of IPv6 tunnels. Set to `false` when routable addresses are being used across the tunnel. Usually when passing an additional subnet.
+**yggdrasil-interface**
+*Server Only*
 
-Default: *true*
-Values: `true` , `false`
+Defines the yggdrasil interface.
```
makeworld-the-better-one:
```suggestion
Defines the Yggdrasil interface.
```

docs/MODULES.md
```diff
@@ -157,63 +157,73 @@ To use this module you must have it installed. You can check to see if the file
...
+Default: `ygg0`
+Value: `Yggdrasil interface on the system, usually ygg0 or tun0`
 
-**SUBNET4**
-*Server only*
+**exit-interface**
+*Server Only*
 
-Defines ip addresses to add to route table that will be routed over through yggdrasil. Must match ips in `yggdrasil.iptunnel.server`
+Defines the exit interface that is connected to the internet.
```
makeworld-the-better-one:
```suggestion
Defines the exit interface that is connected to the Internet.
```

docs/MODULES.md
```diff
@@ -157,63 +157,73 @@ To use this module you must have it installed. You can check to see if the file
...
+  prefix 20xx:xxxx:xxxx::/64
+  {
+    AdvOnLink on;
+    AdvAutonomous on;
+  }
+};
+```
 
-Defines the interface connected to the internet.
+If you installed hostapd you will have this already configured with a prefix.  Simply add the new prefix under the existing one.
```
makeworld-the-better-one: Do they need a comma or some sort of differentiator?

darkdrgn2k:
> It should be in `functions.sh`. 

Functions.sh is used in the install script.  The config include file will actually need to be a module if we will be using it across all the different config durring run time.

docs/MODULES.md
darkdrgn2k: ; - added

scripts/yggdrasil-iptunnel/install
```diff
@@ -6,6 +6,7 @@ BASE_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
 # Dependencies
 sudo apt-get -y install confget
+sudo cp -f ../confset.sh /usr/local/bin/confset.sh
```
makeworld-the-better-one: This should be part of the generic install script, if you're going to leave the file as `scripts/confset.sh`. If you want to make it a module I'm fine with that too, but the way I'm suggesting seems easier.

scripts/confset.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+function confset {
```
makeworld-the-better-one: If this script is going to be moved to `/usr/local/bin`, why not make it a command (instead of this function here) with the name `confset`, so that instead of running `source /usr/local/bin/confset.sh`, you can just run `confset some thing` in the script?

scripts/yggdrasil-iptunnel/yggdrasil-setup
```diff
@@ -4,68 +4,42 @@
...
-    res=$?
-    if [[ "$res" == "1" ]]; then
-        # If file is missing the key it is added
-        sed -i "s/\[$section]/\[$section]\n$key=$value/" "$file"
-    else
-    # Otherwise change it
-        sed -i "/^\[$section]/,/^\[/{s/^$key=.*/$key=$value/;}" "$file"
-    fi
-}
+source /usr/local/bin/config.inc
```
makeworld-the-better-one:
```suggestion
source /usr/local/bin/confset.sh
```

scripts/yggdrasil-iptunnel/install
darkdrgn2k: agreaed

#389

PR #381, opened 2019-04-14
Move node.js into a seperate module ref #374

### LINKED ISSUE #374: move NodeJS into module
https://github.com/tomeshnet/prototype-cjdns-pi/pull/316#discussion_r272845972

cjdns module is the only one that installs NODE.JS but is used by other modules

darkdrgn2k:
Move node.js into a separate module closes  #374

### DIFF
--- a/scripts/cjdns/install
+++ b/scripts/cjdns/install
@@ -5,6 +5,9 @@ TAG_CJDNS=d2e55d58548d83940482fe1bbbe1fd36f7f1b4ef
 # Common functions
 source functions.sh
 
+# Install node.js
+source "$BASE_DIR/../shared/node.js/install"
+
 # Default build script for CJDNS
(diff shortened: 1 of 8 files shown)


darkdrgn2k:
> Happy to see this a module! The [latest LTS Node.js version](https://nodejs.org/en/download/) is 10.15.3 however, and so this script should be updated to use that version.

Not until its tested

To much version hell with JS! #387

makeworld-the-better-one:
Alright, so I'd say let's test with the newest LTS before making it a module and merging this.

darkdrgn2k:
I need the structure to continue DEV work to prevent having to resolve conflicts later. I have merged, tested, fixed and then branched out to new PR #392. Tagged you in it.

Already bumped to 10.x in module

Please proceed and thanks

PR #386, opened 2019-04-19
Netcat fix

makeworld-the-better-one:
@dasanchez and I had an issue with netcat working one meeting, I did some research today, and figured it was [this](https://unix.stackexchange.com/questions/457670/netcat-how-to-listen-on-a-tcp-port-using-ipv6-address). This PR removes potential different versions of netcat, and installs the reccomended one that supports IPv6, `netcat-openbsd`. I have confirmed this is in the Raspbian and Debian repositories.

### DIFF
--- a/scripts/install2
+++ b/scripts/install2
@@ -84,7 +84,7 @@ echo "auto lo" | sudo tee /etc/network/interfaces.d/lo > /dev/null
 echo "iface lo inet loopback" | sudo tee --append /etc/network/interfaces.d/lo >> /dev/null
 
 # Replace netcat with IPv6 supported one, see https://unix.stackexchange.com/questions/457670/netcat-how-to-listen-on-a-tcp-port-using-ipv6-address
-sudo apt-get remove -y netcat-traditional netcat6 ncat netcat
+sudo apt-get remove -y netcat-traditional netcat6 ncat netcat || true
 sudo apt-get install -y netcat-openbsd
 
 # Detect missing /sbin from $PATH variable on Debian distros, and add it
--- a/scripts/install2
+++ b/scripts/install2
@@ -84,8 +84,17 @@ echo "auto lo" | sudo tee /etc/network/interfaces.d/lo > /dev/null
 echo "iface lo inet loopback" | sudo tee --append /etc/network/interfaces.d/lo >> /dev/null
 
 # Replace netcat with IPv6 supported one, see https://unix.stackexchange.com/questions/457670/netcat-how-to-listen-on-a-tcp-port-using-ipv6-address
-sudo apt-get remove -y netcat-traditional netcat6 ncat netcat || true
-sudo apt-get install -y netcat-openbsd
+read -p "Replace any IPv4-only netcat software with one that also supports IPv6 (Y/n)?" -n 1 -r
+echo ""
+if [[ $REPLY =~ ^[Nn]$ ]]; then
+    echo -e "The current software will not be removed."
+else
+    sudo apt-get remove -y netcat-traditional || true
+    sudo apt-get remove -y netcat6 || true
+    sudo apt-get remove -y ncat || true
+    sudo apt-get remove -y netcat || true
+    sudo apt-get install -y netcat-openbsd
+fi
 
 # Detect missing /sbin from $PATH variable on Debian distros, and add it
 if ! echo "$PATH" | grep -q "/sbin" ; then
--- a/scripts/install2
+++ b/scripts/install2
@@ -83,6 +83,10 @@ echo iface eth0 inet dhcp | sudo tee --append /etc/network/interfaces.d/eth0 >> 
 echo "auto lo" | sudo tee /etc/network/interfaces.d/lo > /dev/null
 echo "iface lo inet loopback" | sudo tee --append /etc/network/interfaces.d/lo >> /dev/null
 
+# Replace netcat with IPv6 supported one, see https://unix.stackexchange.com/questions/457670/netcat-how-to-listen-on-a-tcp-port-using-ipv6-address
+sudo apt-get remove -y netcat-traditional netcat6 ncat netcat
+sudo apt-get install -y netcat-openbsd
+
 # Detect missing /sbin from $PATH variable on Debian distros, and add it
 if ! echo "$PATH" | grep -q "/sbin" ; then
     # Current environment


darkdrgn2k:
> I have confirmed this is in the Raspbian and Debian repositories.

Armbian?

makeworld-the-better-one:
I've confirmed Ubuntu contains this pkg, as you can see in your comment. Armbian uses the Debian and Ubuntu pkg repositories for everything other than kernels, and so it will work there as well.

As for placement, netcat comes installed already on most of these boards, but has unexpected behaviour like not working with IPv6. I don't think it should be a contrib file, because it's not adding software, but just fixing software that's already going to be on most nodes. I thought about having it as a part of extra-tools, but it's not additional software. @dasanchez and I already had netcat installed because it comes by default, we expected it to work, and it didn't. I see this code as fixing that small UX problem, so that other users don't have errors like this. They may not want to install *extra* tools, but will want the ones already there to work.

darkdrgn2k:
>  I don't think it should be a contrib file, because it's not adding software, but just fixing software that's already going to be on most nodes

I would disagree. You are in fact rather replacing software, not fixing it.

Man page for NC that comes with debian/ubuntu
https://manpages.debian.org/stretch/netcat-traditional/nc.1.en.html

netcat works as designed - it does not support ipv6

`nc -l -p port [-options] [hostname] [port] `

A different `netcat` that does  support `IPv6` is a completely different program with a completely different set of parameters

Man page for NC 
https://linux.die.net/man/1/nc

`nc [-46DdhklnrStUuvzC] [-i interval] [-p source_port] [-s source_ip_address] [-T ToS] [-w timeout] [-X proxy_protocol] [-x proxy_address[:port]] [hostname] [port[s]]`

It is NOT "backwards" compatible (if you consider this an upgrade)

Doing `nc -l -p 12345` on the first wont would NOT work on the second one. In fact the man page cautions of this

> -p source_port
> ...It is an error to use this option in conjunction with the -l option. 

Just like `ping` did not support IPv6 we had to use `ping6`
and `route` does not support IPv6 you have to use `ip -6 route`

If they called it `nc6` I wouldn't  have a problem with including it
If it was a drop in upgrade I wouldn't have a problem (ie backwards compatible)

But because its not backwards compatible, I want to be careful in changing how things works.  This is how assumptions in dependencies break systems.

> I see this code as fixing that small UX problem

If this is a  UX problem i think its out of scope 
It does not fix a UX of the stack as a whole (maybe 
It fixes your assumption that the stock `netcat` should work with IPv6

> Just Work
If you got this far and you want to play with netcat i dont think we should spoon feed people. Throw it in an FAQ if you want.

If after all this you still think its good to go, then ill bow to the will of the people.

makeworld-the-better-one:
> Doing `nc -l -p 12345` on the first wont would NOT work on the second one. In fact the man page cautions of this

You are looking at the wrong manpage. The netcat-openbsd package manpage is [here](https://man.openbsd.org/OpenBSD-current/man1/nc.1), and a command like `nc -l -p 12345` will work fine.

> If they called it `nc6` I wouldn't have a problem with including it

There is an nc6 pkg, and maybe that's what should be used instead.

> But because its not backwards compatible, I want to be careful in changing how things works. This is how assumptions in dependencies break systems.

This is a good point, and I guess all I wanted was for it to just support IPv6, but even looking at the correct manpage, many options are different, it's just simpler ones like `-l -p 123` that are the same. I will look into nc6 and make sure it's good, and then maybe I can change the code in this PR and I'll add something to the docs that will mention it.

makeworld-the-better-one:
So it appears that netcat6 is available on Raspbian, but only on Jessie for Debian (oldstable), and Sid (unstable), but not the main version at this time, Stretch. That rules it out for me.

At this point we can either replace it with a non-compatible version, or not replace it at all, and direct people to socat and netcat-openbsd in the docs.

darkdrgn2k:
Since the version is backwards compatible im less worries about the reprecautions.

My only concern is scope.  I still think bundeling it into a module would work best even for a long term roadmap

For for the time being I'm happy to leave is as you described in the PR and deal with thoes issues later.

makeworld-the-better-one:
@darkdrgn2k 
> Since the version is backwards compatible im less worries about the reprecautions.

It isn't compatible, it's just that the most common options like `-l` and `-p` still are the same. I don't know what to make of that. The common usage will still work, but more complex things will fail, if they assume the netcat type.

@dasanchez 
> I'd much prefer to make this optional (like Grafana/Prometheus) or even just capture the netcat/socat options in some as-of-yet non-existing FAQ/Tips and Tricks page.

They will be captured when you run `man socat`, or `man nc`, is that enough?

> At the very least, I would like to notify the user that we are removing the stock netcat.

Fair enough, I can add that.

>  Is there another any other package we're removing (whether it gets replaced or not) without alerting the user?

Network Manager only, as far as my searching tells me.

dasanchez:
> > I'd much prefer to make this optional (like Grafana/Prometheus) or even just capture the netcat/socat options in some as-of-yet non-existing FAQ/Tips and Tricks page.
> 
> They will be captured when you run `man socat`, or `man nc`, is that enough?
>
I meant that I'd be satisfied with just telling people "stock netcat is no good for IPV6, but you can do this instead" outside of this software installation rather than forcing the change. I feel like it's somewhat out of scope for what this installation is meant to accomplish

> > At the very least, I would like to notify the user that we are removing the stock netcat.
> 
> Fair enough, I can add that. But won't it be lost in the stream of text of the install?
>
It could be listed next to a `y/[n]` prompt.

> > Is there another any other package we're removing (whether it gets replaced or not) without alerting the user?
> 
> Network Manager only, as far as my searching tells me.
>
This helps. I may wrong, but I feel like Network Manager is a bigger deal than netcat, so at least this change seems consistent with how we treat pre-existing packages. I'll approve now.

makeworld-the-better-one:
> It could be listed next to a `y/[n]` prompt

Ohhh. That would also solve your informing the users problem. I'll add that later, before merging. 

> I may wrong, but I feel like Network Manager is a bigger deal than netcat, so at least this change seems consistent with how we treat pre-existing packages.

Network Manager has no usecase in our stack, and so it was just taking up space. It's a bit different then netcat, and I'd say less important.

darkdrgn2k:
I think an FAQ is a good place to start for sure.  It already includes interesting tidbits bout weird stuff we found.

darkdrgn2k:
> Network Manager has no usecase in our stack, and so it was just taking up space. It's a bit different then netcat, and I'd say less important.

This statement is incorrect. It wasn't dropped because we simply "dont need it". I won't go into details in this thread why network manager was removed but in summary it has a history of unpredictable behaviors and causing problems with traditional networks.  And it seems that at least one distro already dropped it.

This "cleanup" was used to normalize the operating system.

darkdrgn2k:
> Ohhh. That would also solve your informing the users problem. I'll add that later, before merging.

Add it up in the normalization of the os section at least please.

makeworld-the-better-one:
The plot thickens: My RPi appeared to have both `netcat-traditional` and `netcat-openbsd` installed, but the manpages always brought up the BSD one, and it seems like the BSD one is being used. This will not be the case for all machines however, and so I still feel good about removing it. I've added a question to the user.

PR #388, opened 2019-04-19
IPFS update to v0.4.20

makeworld-the-better-one:
See the [CHANGELOG](https://github.com/ipfs/go-ipfs/blob/v0.4.20/CHANGELOG.md#0420-2019-04-16) here. These changes seem to all be under the hood, and won't require changes in our scripts.

Notably, `ipfs swarm connect` now
> marks the newly created connection as "important". This should ensure that the connection manager won't come along later and close the connection if it doesn't think it's being used.

Uploading and finding directories should now be more efficient and faster.

### DIFF
--- a/scripts/ipfs/install
+++ b/scripts/ipfs/install
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 # shellcheck disable=SC1091
-true
 
 set -e
 
-GO_IPFS_VERSION=v0.4.19
+GO_IPFS_VERSION="v0.4.20"
+
 ARCH="$(uname -m)" 
 case "$ARCH" in
   x86_64)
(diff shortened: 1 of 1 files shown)


scripts/ipfs/install
```diff
@@ -1,10 +1,10 @@
 #!/usr/bin/env bash
 # shellcheck disable=SC1091
-true
```
darkdrgn2k: Why was this removed.

Its required to disable sc1091 as part of shellcheck

scripts/ipfs/install
```diff
@@ -1,10 +1,10 @@
...
 
 set -e
 
-GO_IPFS_VERSION=v0.4.19
+GO_IPFS_VERSION="v0.4.20"
```
darkdrgn2k: If tested im good with it

PR #390, opened 2019-04-19
[WIP] - New release v0.5.0-rc.1 to master

darkdrgn2k:
- Yggdrasil IPTunnel now does not change config files
- Yggdrasil IPtunnel drop in service adjustment
- Yggdrsail IPTunnel supports IPv6 and routed IPv6
- Yggdrsail version bump
- Support for x86 and x64
- Profile selection menu format changed in Dialog
- Rewritten and simplified board detection
- IPFS Improvments #284
- CJDNS now module
- Prototype can run without CJDNS now
- Better docs
- Grafan database now can be removed when uninstalling
- Moved network config to interface.d model
- Removed Network Manager
- Added Modules.md
- Added confSet function and implemented confget/confset config files
- NodeJS now shared module
- NodeJS version bump
- MESH_NAME now a config
- Added ipv6 netcat option
- Localized  Patch Foo in TOMESH repo to prevent version conflicts and outage
- Added support for PI4
- Added support for Buster
- Fixed ETH0 vs BR0 issue on espressoBIN
- Prometheus version bump
- Raspberry Pi Watch Dog Timer

PR #391, opened 2019-04-19
CHANGELOG.md file and format

makeworld-the-better-one:
This adds a CHANGELOG.md file that follows [this format](https://keepachangelog.com/en/1.0.0/). It is the same format Yggdrasil uses, and one of the only changelog references on the Web.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,48 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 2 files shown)


makeworld-the-better-one:
We can copy from the file when a release happens, and update the Unreleased section of the file as new PRs are made in `develop`.

makeworld-the-better-one:
@darkdrgn2k Could you add some new stuff to this file, following the existing format. I know there have been new changes since I last committed, I can see them in the commits of #390.

PR #393, opened 2019-04-21
Move confset to seperate binary instead of function

darkdrgn2k:


scripts/shared/confset/uninstall
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# Dependencies
+sudo apt-get -y install confget
```
makeworld-the-better-one:
```suggestion
sudo apt-get -y remove confget
```

PR #399, opened 2019-05-05
Update mesh-point

darkdrgn2k:
Mesh point interfaces sometimes do not enter "RUNNING" mode. This prevents some services like Yggdrasil to peer.

```
wlan1: flags=4099<UP,BROADCAST,MULTICAST>  mtu 1500
```

Seems that wpa_supplicant is interfering with this working on some os/wifi device combinations.

Patch will kill wpa_supplicant if running and not access point defined, but better solution required.

Does not to be a good one at the moment - Refrence
https://serverfault.com/questions/869857/systemd-how-to-selectively-disable-wpa-supplicant-for-a-specific-wlan-interface/880575

### DIFF
--- a/scripts/mesh-point/mesh-point
+++ b/scripts/mesh-point/mesh-point
@@ -21,6 +21,13 @@ set -e
 # Set wireless regulatory domain
 sudo iw reg set CA
 
+# Kill wpa_supplicant if running while no AP active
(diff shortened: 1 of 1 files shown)


darkdrgn2k:
Issue created and mearged

PR #411, opened 2019-05-18
Corrected flow for NetCat replcaemnet module

darkdrgn2k:
replacement of netcat converted to dialog question, and moved to more order after "extra tools" also converted to a module

"text installer" better descibed as "dialog ascii"


scripts/extra-tools-netcat/install
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
makeworld-the-better-one:
```suggestion
#!/usr/bin/env bash
```

scripts/extra-tools-netcat/install
```diff
@@ -0,0 +1,9 @@
...
+
+# Replace netcat with IPv6 supported one
+# see https://unix.stackexchange.com/questions/457670/netcat-how-to-listen-on-a-tcp-port-using-ipv6-address
```
makeworld-the-better-one:
```suggestion
# See https://unix.stackexchange.com/questions/457670/netcat-how-to-listen-on-a-tcp-port-using-ipv6-address
```

scripts/extra-tools-netcat/install
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
```
makeworld-the-better-one:
```suggestion
set -e
```
Should this be here?

scripts/install2
```diff
@@ -291,6 +278,7 @@ if [ "$WITH_PROMETHEUS_SERVER" == true ]; then
     askModule "WITH_GRAFANA" "Grafana"
 fi
 askModule "WITH_EXTRA_TOOLS" "Non-Essential Tools Useful for Network Analysis"
+askModule "WITH_EXTRA_NETCAT" "Replacement netcat that supports IPv6"
```
makeworld-the-better-one:
```suggestion
askModule "WITH_EXTRA_NETCAT" "Change netcat to a version that supports IPv6"
```

darkdrgn2k: Prompt prepends "INSTALL" in front of text.

`Install change netcat` does not read well

PR #414, opened 2019-05-18
Contrib folder explanation

### LINKED ISSUE #410: [Suggestion] README.md on /contrib folder

makeworld-the-better-one:
Closes #410

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ The software also supports Debian running on x86 and x64 boards, and many board 
 
 During the installation, you may be able to pick a profile, or choose between many modules. To learn what each module is for, look at [MODULES.md](./docs/MODULES.md). This is important for the installation.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,6 +44,8 @@ The software also supports Debian running on x86 and x64 boards, and many board
 
 During the installation, you may be able to pick a profile, or choose between many modules. To learn what each module is for, look at [MODULES.md](./docs/MODULES.md). This is important for the installation.
 
+There are some more experimental, 3rd party, or not fully supported modules in the [contrib](./contrib) folder for you to explore. Run those scripts with caution.
```
darkdrgn2k: Contrib folders are not normally defined as they are a standard in many projects

What is the contrib folder:

> It is for software that has been contributed to the project, but which might not actually be maintained by the core developers. Naming it "contrib" or "Contrib" is a long established convention.

In our case many of the scripts augment existing modules for specific task, and may need guidance to use. At present they are not documents at all.

makeworld-the-better-one: I will change my definition to the one you quoted.

PR #424, opened 2019-06-16
[wip] - Babeld - Debian Package POC 

darkdrgn2k:
Proof of Concept debian packages WIP
Babeld installed from the https://github.com/darkdrgn2k/packages/ repo 

The configuration after would be rolled into a  "babeld-tomesh" package/


makeworld-the-better-one:
I don't feel good about having a module that uses debian packages from a non-tomesh repo, and I thought we were waiting until v0.6 to move to Debian packages anyway. If you want to add it as a `contrib/` pkg I'd be okay with that though.

PR #434, opened 2019-07-19
Patch wpa_supplicant for Raspbian

### LINKED ISSUE #403: WPA_SUPPLICANT - better way to deal with it
Work around made for this issue on #399 There is a similar one in ADHOC. Perhaps it be nice to have a better solution to isolate single interfaces

---  issue ----

Mesh point interfaces sometimes do not enter "RUNNING" mode. This prevents some services like Yggdrasil to peer.

```
wlan1: flags=4099<UP,BROADCAST,MULTICAST>  mtu 1500
```

Seems that wpa_supplicant is interfering with this working on some os/wifi device combinations.

Patch will kill wpa_supplicant if running and not access point defined, but better solution required.

Does not to be a good one at the moment - Refrence
https://serverfault.com/questions/869857/systemd-how-to-selectively-disable-wpa-supplicant-for-a-specific-wlan-interface/880575

darkdrgn2k:
Fixes #403 

Patch reads a file `/etc/wpa_devices` for list of interfaces and ignores any interfaces that are not on the list.

### DIFF
--- a/scripts/hostapd/install
+++ b/scripts/hostapd/install
@@ -154,5 +154,5 @@ fi
 
 # Apply wpa_supplicant patch 
 # shellcheck disable=SC2024
-sudo patch -F99 -p2 -N -d / < wpa-patch.diff
+sudo patch -F99 -p2 -N  -d / < wpa-patch.diff
 echo "wlan-ap" | sudo tee /etc/wpa_devices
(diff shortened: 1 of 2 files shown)


darkdrgn2k:
- [ ] Does not work on armbian?

darkdrgn2k:
Arbmian does things differently. This patch is for raspbian only

darkdrgn2k:
:heavy_check_mark: Tested on stock Raspberry Pi 4 raspbian september 2019

Issue #350, opened 2019-03-13
ssb peer announcement over cjdns


darkdrgn2k:
sbot does not listen on ::8008

need to bridge ::8008 and 0.0.0.0:8008

darkdrgn2k:
Forward udp6 to udp4
`socat UDP6-LISTEN:8008,fork,su=nobody UDP4:127.0.0.1:8008`

needs to look if sbot actually picks this up

makeworld-the-better-one:
What's the status on this atm?

darkdrgn2k:
`socat UDP6-LISTEN:8008,fork,su=nobody UDP4:127.0.0.1:8008` may be the solution to re-route ipv6 packets to the ipv4 port

testing needed

PR #435, opened 2019-07-19
socat SSB proxy

darkdrgn2k:
added SOCAT to proxy broadcast packest aimed at ipv6 (cjdns/yggdarsil) packets to ssb's ipv4 port

Ref #350


darkdrgn2k:
# Testing

- Build two nodes with SSB and CJDNS with Mesh point or AdHoc
- Make sure at least one node is NOT connected to the lan (only mesh)
- Node should only be reachable by 

Check if peering occurred
(ie `sbot gossip.peers` )

Issue #322, opened 2019-03-04
Yggdrasill internet peer and ip tunnel conflict


darkdrgn2k:
If yggdrasill is using an interent peer, and set to ipclient it will fail

Reason

1. yggdrasil will establish a connection to the internet peer
1. yggdrasill will establish the iptunnel over to the yggdrasill exit node
1. yggdrasill will run all traffic over the new iptunnle to the exit node
1. yggdrasill will try to send all peering trafic over the exit node
1. yggdrasill peering will fail because its tries to send info over the ip tunnel the is depended on the peer  that is now depended on the iptunnel
1. It will stop working

work around - createa a route over your internet connection to he exit peer

`sudo ip route add 1.1.1.1 gw 2.2.2.2`

where 1.1.1.1 is your internet peer and 2.2.2.2 is your internet gateway

Maybe not action needs to be taken but maybe just document this?

makeworld-the-better-one:
Has this been resolved? I saw documentation about doing exactly this in a recent PR, #377 maybe.

darkdrgn2k:
Aside from note in the docs it is NOT "fixed"

PR #437, opened 2019-07-20
Yggdrasill Gateway+Peer workaround

darkdrgn2k:
ref #322 

Two options to work around bug identified in #322 

Options defined in `/etc/yggdrasil.iptunnel.conf` as "reroutepeers"
```
[general]
reroutepeers=unroutepeers
```
Default - empty no workaround

# reroutepeers=routepeers
Iterate through peers and add them to the systems default gateway
Add default routes over ygg

## Pros
Currently works for all listed peers

## Cons
When gateway changes tunnel collapses 

# reroutepeers=unroutepeers
Instead adding default routes progressively add routes but omitting the peer's ip

# Pros
When ygg interface collapses routes are removed
Since route for peer is never set it always uses the system default (survives roaming wifi etc)

# Cons
Currently works for only the first peer
Fills up the route table with routes

# Additional Notes

- when JQ installed on system, it will read the yggdarsill config file for peers. 
- if JQ is not installed uses getPeers which means the peer must already be connected
- Cant parse hostnames (assumes ips)
- If/When Hostnames are used, DNS resolve may be an issue (round robin dns)


darkdrgn2k:
# TESTING

- Install yggdarsill 
- Peer TCP to another node
- Make sure that TCP is the only path to the exit node
- Configure IPTunnel.client 
- Report success and failures

docs/MODULES.md
```diff
@@ -303,18 +303,56 @@ or with IPv6 and subnet
 
 ##### IPTunnel Client and Internet Peer
 
-If you wish to run an IPTunnel client on the same node as as an Internet peer, you will need to create a route to that node over the Internet connection available.
+If you wish to run an IPTunnel client on the same node as as an Internet peer, you will need to use a method of rerouting the path to the Yggdrasill peers Internet connection available instead over yggdrasil.
```
benhylau:
```suggestion
If you wish to run an IPTunnel client on the same node as an Internet peer, you will need to ensure the path to the Yggdrasil peer goes over the Internet connection instead of using the tunnel.
```

docs/MODULES.md
```diff
@@ -303,18 +303,56 @@ or with IPv6 and subnet
...
 
-This must be done because since all Internet traffic will be redirected over the tunnel and a route is not created to the peer, Yggdrasil peering will try to route over the tunnel. Since the tunnel depends on the peer, the peer will not function, which will collapse the tunnel. In other words, the peer tries to feed packets over the tunnel, that tries to feed packets to the peer, causing both to fail.
+This must be done because since all Internet traffic will be redirected over the tunnel.  If a route is not created to the peer, Yggdrasil peering will try to route over the tunnel. Since the tunnel depends on the peer, the peer will not function, which will collapse the tunnel. In other words, the peer tries to feed packets over the tunnel, that tries to feed packets to the peer, causing both to fail.
```
benhylau:
```suggestion
This must be done because all Internet traffic will be redirected over the tunnel. If a route is not created to the peer, Yggdrasil peering will try to route over the tunnel. Since the tunnel depends on the peer, the peer will not function, which will collapse the tunnel. In other words, the peer tries to feed packets over the tunnel, that tries to feed packets to the peer, causing both to fail.
```

docs/MODULES.md
```diff
@@ -303,18 +303,56 @@ or with IPv6 and subnet
...
-Add the following command to your `/etc/rc.local`
-```
-route add 123.123.123/24 gw 192.168.0.1
-```
-Where:
-`123.123.123` is the IP address of the Yggdrasil peer  
-`192.168.0.1` is the current active gateway to the Internet, ie your router.
+There are two ways of rerouting packets as part of this package.
+
+Method is configed in the `[general]` secion of yggdrasill.iptunnel.config
```
benhylau:
```suggestion
Method is configured in the `[general]` section of yggdrasill.iptunnel.config
```

docs/MODULES.md
```diff
@@ -303,18 +303,56 @@ or with IPv6 and subnet
...
+
+**Cons**
+
+Currently works for only the first peer
+Fills up the route table with routes
+
+###### Additional Notes
 
+- Does not yet work for IPv6 peers
+- When JQ is installed on system, it will read the yggdarsill config file for peers
```
benhylau: Do a global spell-check of yggdrasil

PR #440, opened 2019-08-19
Update IPFS

makeworld-the-better-one:
[Changelog](https://github.com/ipfs/go-ipfs/blob/master/CHANGELOG.md#0422-2019-08-06)

This is only a patch release, fixing some important bugs. No changes should be necessary on our side.


kylerchin:
UPDATE IT!

PR #441, opened 2019-08-19
Update Yggdrasil to 0.3.11

makeworld-the-better-one:
[Changelog](https://github.com/yggdrasil-network/yggdrasil-go/blob/master/CHANGELOG.md#038---2019-08-21) - scroll down and look at 0.3.6 and 0.3.7 changes as well.

Because the last version was 0.3.5, there should probably be some testing to make sure everything works, but it doesn't look like there have been changes that would affect us. @darkdrgn2k you might know better?


makeworld-the-better-one:
I will be updating this later today for the recent v0.3.8 release.

makeworld-the-better-one:
From the changelog:
- Crypto-key routing options have been renamed for clarity
  - `IPv4Sources` is now named `IPv4LocalSubnets`
  - `IPv6Sources` is now named `IPv6LocalSubnets`
  - `IPv4Destinations` is now named `IPv4RemoteSubnets`
  - `IPv6Destinations` is now named `IPv6RemoteSubnets`
  - The old option names will continue to be accepted by the configuration parser for now but may not be indefinitely

PR #442, opened 2019-08-22
Syntax issue (capitalization)

darkdrgn2k:
Corrected variable incorrectly commited by @kylerchin PR
should be UPPERCASE committed as MIXED case


kylerchin:
eggcellent!

PR #458, opened 2019-10-27
Allow non-onboard wifi as AP

darkdrgn2k:
AP using onboard WiFi now optional
If not selected Extenral card can be used by pinning the MAC

## TODO

- [x] Fix `askSelection` - Does not like passing \n
- [x] Fix `askSelection` -  May not like return numbers 
- [x] Add `hostapd` as install-able option globally

- [x] Test on Rasbian
- [ ] Test on Armbian

## Stretch
- [ ] Convert to check all network interfaces instead of just wlan*
    - [ ]  Check for wireless folder to check if net is wireless  (reference mesh-point script)
- [x] Pin by driver instead of MAC
- [ ] Pin to position


darkdrgn2k:
`askSelection` works fine, was using the wrong branch to test.

darkdrgn2k:
Stretch goal - Position pining

`ENV{ID_PATH}` - possibly a variable for rule
`udevadm info /sys/class/net/wlan-ap` - List all udev attributes

PR #465, opened 2019-11-25
Ygg Version 0.3.12

darkdrgn2k:
new feature should eliminate the need for unroute and route peering with iptunnel


darkdrgn2k:
:heavy_check_mark: Works great as an gateway and client node. Remmeber to use InterfacePeers instead of Peers to config peers.

PR #483, opened 2020-04-11
Update NodeJS to latest LTS (12.16.2)

makew0rld:
This is a small PR, but the biggest change is that `armv6` is left behind. NodeJS no longer supports it, and so I've set it up to download the 10.x LTS instead. Should we be compiling our own versions of NodeJS for armv6 and 32-bit arches, which are stuck on the 8.x LTS? I opened #484 to talk about that. For now I think this can be merged.

### DIFF
--- a/scripts/shared/node.js/install
+++ b/scripts/shared/node.js/install
@@ -2,18 +2,18 @@
 
 set -e
 
-NODEJS_PREFIX=10
-NODEJS_VERSION="$NODEJS_PREFIX.15.3"
+NODEJS_PREFIX=12
+NODEJS_VERSION="$NODEJS_PREFIX.16.2"
 
 # Check whether it has the newest version - 32-bit OSes need an older one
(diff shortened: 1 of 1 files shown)


makeworld-the-better-one:
This needs to be confirmed to work - that it can successfully install ssb, before merging. I have tried and it failed. I believe this is because `ssb-server` relies on the `sodium-native` pkg, at v2.4.2. I have not confirmed this myself, but it's what @darkdrgn2k said. As can be seen in [this issue](https://github.com/sodium-friends/sodium-native/issues/95), only in v2.4.**3** and above did `sodium-native` support building with Node v12. As such, using this newer Node version causes the command `npm install ssb-server` to fail.

makeworld-the-better-one:
On the [ssb-server README](https://github.com/ssbc/ssb-server/blob/master/README.md) they say to install Node v10 explicitly. We could use a solution like `nvm` to manage multiple Node versions, but unless some other part of the code needs a higher version than maybe this PR can stay unmerged.

darkdrgn2k:
Moved to 14

